PR #11, opened 2022-02-12
feat: create camp info csv + update mongoose

SinghHarman286:
## Notion ticket link

[Camp Info CSVs](https://www.notion.so/uwblueprintexecs/Task-Board-38b49f83f6ce4b07ae188dabb37631e8?p=f0cb9e7ec9eb4f4081955cefca07c111)

## Implementation description
* Created campRoute.ts, and added camp/csv/:id endpoint
* Created interface/campService.ts, implementations/campService.ts for csv generation logic
* Updated mongoose to latest version

## Steps to test
1. You can grab any id from camp model, and make request to http://localhost:5000/camp/csv/<id>

* Had to change campers type in Camp.model.ts from Schema.Types.ObjectId[] to PopulatedDoc<CamperDTO>[] because upon populating the campers (array of object ids), typescript treat it as array of object ids and not CamperDTO. So, I had to choose PopulatedDoc type of mongoose (https://mongoosejs.com/docs/typescript/populate.html). I am not sure if this is the best approach (still looking and if found anything, I'll change it). Also, if we do go with PopulatedDoc, then we'll need to change all other types in other files too.
*  Currently, when there are no campers, I am returning an empty string. So, want to make to sure if this should be the desired output.

## Checklist
- [x ] My PR name is descriptive and in imperative tense
- [x ] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [x ] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR


jjjding2:
For the PopulatedDoc type, maybe this helps? 
https://stackoverflow.com/questions/47923403/typescript-how-to-define-model-in-combination-with-using-mongoose-populate

backend/typescript/types.ts
```diff
@@ -37,6 +37,8 @@ export type CamperDTO = {
   chargeId: number;
 };
 
+export type CamperCSVInfoDTO = Omit<CamperDTO, "camps" | "id">;
```
annechung, line 40: update camps -> camp since this is being changed on Camper model
```suggestion
export type CamperCSVInfoDTO = Omit<CamperDTO, "camp" | "id">;
```

annechung (commented):
awesome work, thanks for looking into the object population stuff :DD let me know if the mongoose version update works!

backend/typescript/rest/campRoutes.ts
```diff
@@ -0,0 +1,19 @@
...
+import { getErrorMessage } from "../utilities/errorUtils";
+
+const campRouter: Router = Router();
+
+const campService: ICampService = new CampService();
+
+campRouter.get("/csv/:id", async (req, res) => {
+  try {
+    const csvString = await campService.generateCampersCSV(req.params.id);
+    res.status(200).json(csvString);
```
annechung, line 13: We might have to set the content type to 
```suggestion
    res.status(200).set('Content-Type', 'text/csv').send(csvString);
```

SinghHarman286: Good call, and sure, I'll update this!

annechung (commented):
thanks for updating the version! few things:

getting this error in the backend container on docker-compose up:
image

also did you update mongoose with yarn or npm? we should use yarn (I think this should update the yarn.lock file)

annechung (approved):
tested and works well - feel free to merge this! thanks for updating mongoose 🚀

PR #18, opened 2022-02-24
feat: update campleader camps

amanda-guo:
## Notion ticket link
[Update User API](https://www.notion.so/uwblueprintexecs/Update-User-API-fc75bd22e9f546c788363b8a4a6f84bb)

## Implementation description
* Changed the UpdateUserDTO to take in an optional `camps` field for CampLeaders
* Added a validator to validate that `camps` is an array of strings in `updateUserDtoValidator`
* Modified `updateUserById` to update camps if the role of the user is CampLeader

## Steps to test
1. On the MongoDB console, choose a user in the `users` table that has a valid `authId` and has a CampLeader role. Make sure to add a KVP for `__t: "CampLeader"` as the discriminator key if one doesn't already exist.
2. Using Postman, do a PUT request to `http://localhost:5000/users/:userId/`. In the "Path Variables" under "Params", add the userId and in the "Body" add the updated parameters in JSON. Note: if adding to `camps`, make sure that you're adding ObjectIds to the array.
3. Submit the request and check that it's successful and that the changes appear on the MongoDB console for the chosen user.
4. **Note:** can also try testing edge cases (updating an Admin should not update the list of camps, can add an empty camps array to a CampLeader, error thrown if not passing an array of ObjectIds as strings to camps) 

* Good code placement
* Any checks that might have been missed in the code
* Any edges cases that might have been missed in testing

## Checklist


annechung (commented):
tested and works as expected! one thing, currently we have to pass in all user fields even if we're just updating one field like the camps field - is it possible to update this so that only the fields that are passed are updated? to do this I think we can update the validator so it only checks fields that are on the request body (ex.` req.body.firstName && validatePrimitive(...)` for every field)

can you try it out and let me know if it works?

annechung (approved):
added one commit for updating the user email only if it's passed in in the request. tested and looks good, awesome job with this :D

PR #20, opened 2022-03-05
feat: registration api capacity check

rchang0501:
## Notion ticket link
[Registration API: capacity check](https://uwblueprintexecs.notion.site/Registration-API-capacity-check-136ccc0ae90f4169a8e37351bb5ad302)

## Implementation description
* Added camp capacity check to create camper service

## Steps to test
1. Make a new POST request at http://localhost:5000/camper/register
2. Set the camp field to a camp that's full --> request should through an error stating "Error: camp is full" 
3. Set the camp field to a camp that's not full --> request should return request body as expected 

* Are the error messages formatted correctly? 
* Is there a more efficient way to do this? 

## Checklist

### DIFF
--- a/backend/typescript/services/implementations/camperService.ts
+++ b/backend/typescript/services/implementations/camperService.ts
@@ -13,6 +13,20 @@ class CamperService implements ICamperService {
     let newCamper: Camper;
     let existingCamp: Camp | null;
     try {
+      existingCamp = await MgCamp.findById(camper.camp);
+
+      if (existingCamp) {
+        const existingCampObj = existingCamp.toObject();
+
+        if (existingCamp.campers.length >= existingCampObj.capacity) {
+          throw new Error(
+            `Error: camp is full. Current number of campers in camp: ${existingCamp.campers.length}. Camp capacity: ${existingCampObj.capacity}.`,
+          );
+        }
+      } else {
+        throw new Error(`Camp ${camper.camp} not found.`);
+      }
+
       newCamper = await MgCamper.create({
         camp: camper.camp,
         registrationDate: camper.registrationDate,
--- a/backend/typescript/services/implementations/camperService.ts
+++ b/backend/typescript/services/implementations/camperService.ts
@@ -16,11 +16,9 @@ class CamperService implements ICamperService {
       existingCamp = await MgCamp.findById(camper.camp);
 
       if (existingCamp) {
-        const existingCampObj = existingCamp.toObject();
-
-        if (existingCamp.campers.length >= existingCampObj.capacity) {
+        if (existingCamp.campers.length >= existingCamp.capacity) {
           throw new Error(
-            `Error: camp is full. Current number of campers in camp: ${existingCamp.campers.length}. Camp capacity: ${existingCampObj.capacity}.`,
+            `Error: camp is full. Current number of campers in camp: ${existingCamp.campers.length}. Camp capacity: ${existingCamp.capacity}.`,
           );
         }
       } else {


backend/typescript/models/user.model.ts
```diff
@@ -26,7 +26,7 @@ const UserSchema: Schema = new Schema({
   },
   email: {
     type: String,
-    required: true,
+    required: false,
```
rchang0501, line 29: I didn't change this value... I fetched and pulled from main first and then made my branch - I made sure my main was up to date as well so I'm unsure why this is here...

backend/typescript/services/implementations/camperService.ts
```diff
@@ -12,7 +12,23 @@ class CamperService implements ICamperService {
...
+
     try {
+      existingCamp = await MgCamp.findById(camper.camp);
+
+      if (existingCamp) {
+        const existingCampObj = existingCamp.toObject();
+
+        if (
+          existingCamp &&
+          existingCamp.campers.length === existingCampObj.capacity
```
jjjding2, line 24: I think we should have 
`existingCamp.campers.length >= existingCampObj.capacity`
to be safe. There could be an edge case where the capacity changes to a lower value or we erroneously add too many campers, and we want to be able to catch that when creating new campers.

rchang0501: done!

backend/typescript/services/implementations/camperService.ts
```diff
@@ -12,7 +12,23 @@ class CamperService implements ICamperService {
...
     let existingCamp: Camp | null;
+
     try {
+      existingCamp = await MgCamp.findById(camper.camp);
+
+      if (existingCamp) {
+        const existingCampObj = existingCamp.toObject();
+
+        if (
+          existingCamp &&
```
amanda-guo, line 23: do we need this second check for `existingCamp` if the check is already done in the if statement above?

backend/typescript/services/implementations/camperService.ts
```diff
@@ -12,7 +12,23 @@ class CamperService implements ICamperService {
...
+          existingCamp.campers.length >= existingCampObj.capacity
+        ) {
+          throw new Error(`Error: camp is full`);
```
amanda-guo, line 26: very nitpicky thing: the error message could be slightly more descriptive, like can print out the camp capacity and the length of campers

backend/typescript/services/implementations/camperService.ts
```diff
@@ -13,6 +13,20 @@ class CamperService implements ICamperService {
     let newCamper: Camper;
     let existingCamp: Camp | null;
     try {
+      existingCamp = await MgCamp.findById(camper.camp);
+
+      if (existingCamp) {
+        const existingCampObj = existingCamp.toObject();
```
annechung, line 19: why did you need this `existingCampObj`, were you getting an error? I just tested your code using only existingCamp and it worked fine - I was getting this error from this line: "Property 'toObject' does not exist on type 'Camp'.ts(2339)"

annechung (commented):
this is a good start but I don't think this completely prevents over-registrations! for example if two people were registering for a camp with one spot left at the same time and they both read the value of the camp before the other person's registration was processed (so campers.length was still under capacity for both users) they would still both be able to register, which shouldn't be possible. we should be validating at the time of the update (since updates are guaranteed to happen one at a time). could you look into whether or not we can validate this on update (like update our model to check for capacity and campers length on update)? also, we wanted to move capacity to the baseCamp model, would that cause problems with a check like this? let me know if you have any questions

backend/typescript/services/implementations/camperService.ts
rchang0501: Yea I couldn't properly access the capacity property unless I cast the existing camp into an object --> now I see the same error as you so I'll remove it

annechung (approved):
looks good, feel free to merge this! will create another task for the other piece :)

PR #21, opened 2022-03-05
feat: create get camp endpoint

SinghHarman286:
## Notion ticket link

[Get Camps](https://www.notion.so/uwblueprintexecs/Tasks-38b49f83f6ce4b07ae188dabb37631e8?p=e8830a0c18d04bb7bebf1a8c23397c4f)

## Implementation description
* fixed baseCamp type in camp.model.ts from `Schema.Types.ObjectId` to `BaseCamp | Schema.Types.ObjectId;` in order to support mongoose populate
* added a GET camp route - /camp
* added getCamps service that returns an array of GetCampDTO objects
* fixed camp csv service

## Steps to test
1. when there are no camps, the API should return an empty array []
2. you can call POST /camp/ to create a camp, and then GET /camp/ to view the camp

Call POST /camp with the following JSON body to create a camp:
`{
    "active": true,
    "ageLower": 15,
    "ageUpper": 30,
    "name": "test camp",
    "description": "description",
    "location": "canada",
    "capacity": 20,
    "startTime": "12:00",
    "endTime": "17:00",
    "dates": ["Sun Mar 13 2022 20:01:14 GMT-0600 (Mountain Daylight Time)"],
    "fee": 25,
    "formQuestions": [{"question": "Q1", "type": "Text", "required": true, "description": "question description"}]
}`

Call GET /camp to get all the camps

Sample Response:

Screenshot 2022-03-13 at 8 26 54 PM
9250baf6fceb.png)

* verify if "get camp" route response contain all the desired fields

## Checklist


backend/typescript/services/implementations/campService.ts
```diff
@@ -1,17 +1,85 @@
...
+          model: MgFormQuestion,
+        },
+      });
+
+      if (!camps) {
+        return [];
+      }
+
+      return camps.map((camp) => {
+        const baseCamp = camp.baseCamp as BaseCamp;
```
jjjding2, line 37: I think there might be a compiler error here?

image

SinghHarman286: hmm, I don't see it from my end. Are you sure you have `baseCamp: BaseCamp | Schema.Types.ObjectId;` and not `baseCamp: Schema.Types.ObjectId;` in your camp.model.ts. That change is included in my commit https://github.com/uwblueprint/focus-on-nature/blob/feat/create-get-camp-endpoint/backend/typescript/models/camp.model.ts#L6

jjjding2: ah you're right sorry – the merge conflict didn't show up in gh desktop..

I was looking through your PR to see what needs to be updated since we're refactoring the baseCamp/Camp model, hope I didn't wake you up

SinghHarman286: okk, no worries. I usually end up sleeping pretty late Friday nights so all good :)

backend/typescript/services/implementations/campService.ts
```diff
@@ -124,13 +210,22 @@ class CampService implements ICampService {
...
+        const { formResponses, ...formObj } = camper;
+        return {
+          ...formResponses,
+          ...formObj,
+        };
+      });
       // grabbing column names
-      const fields = Object.keys(campers[0]);
-      const csvString = await generateCSV({ data: campers, fields });
+      const fields = Object.keys(flattenedCampers[0]);
```
jjjding2, line 227: Is it possible for campers to have different formResponses keys? Wondering if we need to take the union of keys for all campers.

backend/typescript/types.ts
```diff
@@ -83,6 +84,11 @@ export type CampDTO = {
   active: boolean;
 };
 
+export type GetCampDTO = Omit<
+  CampDTO & BaseCampDTO,
+  "baseCamp" | "campers" | "waitlist" | "camps"
```
jjjding2, line 89: Do we need to return "campers" and "waitlist"? Or will this only be required when getting the csv or a specific camp

jjjding2 (commented):
Tested and looks good! Left some comments

backend/typescript/types.ts
SinghHarman286: I remember that I discussed the same question with @annechung, and it was decided that campers and waitlist are not to be returned, but just to confirm again, @annechung do we need "campers" and "waitlist" in get camp service?

annechung: yepp that's right, we only need to get campers and waitlist for specific camps

backend/typescript/services/implementations/campService.ts
SinghHarman286: a camp has a specific set of form questions we are retrieving from, i.e. baseCamp has formQuestions field containing all the questions. By union, did you mean taking union of all campers from all the camps? cc: @annechung

backend/typescript/services/implementations/campService.ts
```diff
@@ -1,17 +1,93 @@
...
+          model: MgFormQuestion,
+        });
+
+      if (!camps) {
+        return [];
+      }
+
+      return camps.map((camp) => {
+        const formQuestionArr = camp.formQuestions as FormQuestion[];
+        const formQuestions = formQuestionArr.map(
```
annechung, line 43: nit: do you need the formQuestionsArr or can you do this?
```suggestion
        const formQuestions = (camp.formQuestions as FormQuestion[]).map(

```

backend/typescript/services/implementations/campService.ts
```diff
@@ -1,17 +1,93 @@
...
+          id: camp.id,
+          ageLower: camp.ageLower,
+          ageUpper: camp.ageUpper,
+          capacity: camp.capacity,
+          name: camp.name,
+          description: camp.description,
+          location: camp.location,
+          fee: camp.fee,
+          formQuestions,
+          campSessions,
```
annechung, line 82: can we actually update this to return campers and waitlist fields for each campSession, but just the numbers of people on each list? so the frontend will have the number of people on the campers and waitlist - the designs require thiss https://www.figma.com/file/uRMwPpIaSgsMGdbJzZziMA/FON-%2F-Management-Experience?node-id=407%3A7960

backend/typescript/services/implementations/campService.ts
```diff
@@ -26,16 +102,34 @@ class CampService implements ICampService {
...
-    } catch (error: unknown) {
-      Logger.error(
-        `Failed to get entities. Reason = ${getErrorMessage(error)}`,
+      return await Promise.all(
+        campers.map(async (camper) => {
+          const { formResponses } = camper;
+          const formResponseObject: { [key: string]: string } = {};
+          const formQuestionsPromise = Array.from(formResponses.keys()).map(
+            (questionId) => {
+              return MgFormQuestion.findById(questionId, "question");
```
annechung, line 111: hmm I think we can get all of the formQuestions at once? to reduce individual calls to the db - we can do something like` MgFormQuestion.find({'_id': {$in: [all ids here]}})`? https://stackoverflow.com/questions/8303900/mongodb-mongoose-findmany-find-all-documents-with-ids-listed-in-array

backend/typescript/services/implementations/campService.ts
```diff
@@ -26,16 +102,34 @@ class CampService implements ICampService {
...
+            },
+          );
+          const formQuestions = await Promise.all(formQuestionsPromise);
+          formQuestions.forEach((formQuestion) => {
+            if (formQuestion) {
```
annechung, line 116: do you need this if statement?

backend/typescript/services/implementations/campService.ts
annechung: also is it possible to only grab all of the questions once? so outside of the campers map function, to avoid getting all the questions multiple times. you can maybe do a union of all question ids of all campers, get all of the question objects from the db (making a map of question id to question object), then for each camper create the formResponse using the map we created earlier. what are you thoughts, do you think this would be too complicated?

backend/typescript/services/implementations/campService.ts
```diff
@@ -140,9 +234,17 @@ class CampService implements ICampService {
         // if there are no campers, we return an empty string
         return "";
       }
+
+      const flattenedCampers = campers.map((camper) => {
+        const { formResponses, ...formObj } = camper;
+        return {
+          ...formResponses,
+          ...formObj,
+        };
```
annechung, line 243: is it possible to do this?
```suggestion
        return {
          ...camper.formResponses,
          ...camper.formObj,
        };
```

backend/typescript/services/implementations/campService.ts
annechung: discussed a bit offline but no not a union of all questions from all camps, but from all campers i the same camp!

annechung (commented):
good start! just a few nits and one small update to GetCampDTO we want to make

backend/typescript/services/implementations/campService.ts
SinghHarman286: thats a very good catch! I'll include it in my PR!

backend/typescript/services/implementations/campService.ts
SinghHarman286: camper does not have "formObj" field,

`const {formResponses, ...formObj} = campers;` // here I am getting formResponses field from camper object, and everything else is encapsulated in "formObj"

does it make sense? did I understand your question correctly?

backend/typescript/services/implementations/campService.ts
SinghHarman286: sure, added! In design dock, I see that they have referred to it as "registrations"  and "waitlist", so the new field include "registrations" and "waitlists" containing number of campers in each of them.

backend/typescript/services/implementations/campService.ts
```diff
@@ -1,17 +1,85 @@
...
+              description,
+              options,
+            } = formQuestion;
+            let result: FormQuestionDTO = { id, type, question, required };
+            if (description) {
+              result = { ...result, description };
+            }
+            if (options) {
+              result = { ...result, options };
+            }
```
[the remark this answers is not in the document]
SinghHarman286: nice catch! makes much more sense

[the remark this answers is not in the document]
SinghHarman286: I think that unpacking formQuestion results in mongoose fields being added: for eg. `"$__":{"strictMode":true,"selected":{},"getters":{},...` etc. Alternatively, we can just use this

`return formQuestion;` but this will result in __v and _id being added. For example:

`{
options: [],
_id: 6236dab5a84ccbbe42d813dd,
type: 'Multiselect',
question: 'Q1',
required: true,
description: 'question description',
__v: 0
}`

backend/typescript/types.ts
```diff
@@ -70,15 +84,12 @@ export type CampDTO = {
...
-  dates: string[];
-  startTime: string;
-  endTime: string;
-  active: boolean;
+export type GetCampDTO = Omit<CampDTO, "campSessions" | "formQuestions"> & {
+  formQuestions: FormQuestionDTO[];
+  campSessions: (Omit<
+    CampSessionDTO,
+    "id" | "camp" | "campers" | "waitlist"
+  > & { registerations: number; waitlists: number })[];
```
annechung, line 92: think this naming makes it a bit clearer, what do you thinkk 
```suggestion
  > & { numCampers: number; numWaitlist: number })[];
```

backend/typescript/services/implementations/campService.ts
annechung: oh whoops i thought i responded to this - it should be just the union of all campers from a specific campsession

backend/typescript/models/baseCamp.model.ts
```diff
@@ -0,0 +1,59 @@
+import { Schema, Document, model } from "mongoose";
+import { FormQuestion } from "./formQuestion.model";
+
+export interface BaseCamp extends Document {
```
annechung, line 4: can we delete this filee

backend/typescript/models/camper.model.ts
```diff
@@ -6,7 +6,7 @@ export interface Camper extends Document {
   formResponses: Map<string, string>;
   registrationDate: Date;
   hasPaid: boolean;
-  chargeId: string;
+  chargeId: number;
```
annechung, line 9: chargeId should be string type!

backend/typescript/services/implementations/campService.ts
```diff
@@ -1,41 +1,144 @@
...
-  async getCampersByCampId(campId: string): Promise<CamperCSVInfoDTO[]> {
+  async getCamps(): Promise<GetCampDTO[]> {
     try {
-      const camp: CampSession | null = await MgCampSession.findById(
-        campId,
+      const camps: Camp[] | null = await MgCamp.find({})
+        .populate({
+          path: "campSessions",
+          model: MgCampSession,
+          populate: {
```
annechung, line 26: do we need to populate campers?

backend/typescript/types.ts
```diff
@@ -34,11 +34,14 @@ export type CamperDTO = {
   campSession: string;
   registrationDate: Date;
   hasPaid: boolean;
-  chargeId: string;
+  chargeId: number;
```
annechung, line 37:
```suggestion
  chargeId: string;
```

backend/typescript/services/implementations/campService.ts
```diff
@@ -1,41 +1,144 @@
...
+            };
+          },
+        );
+
+        const campSessions = (camp.campSessions as CampSession[]).map(
+          (campSession) => ({
+            dates: campSession.dates.map((date) => date.toString()),
+            startTime: campSession.startTime,
+            endTime: campSession.endTime,
+            registerations: campSession.campers.length,
```
annechung, line 59:
```suggestion
            registrations: campSession.campers.length,
```

backend/typescript/services/implementations/campService.ts
```diff
@@ -1,41 +1,140 @@
...
+            const answer = formResponses.get(questionId);
+            formResponseObject[question] = answer as string;
+          });
+
+          return {
+            formResponses: formResponseObject,
+            registrationDate: camper.registrationDate,
+            hasPaid: camper.hasPaid,
+            chargeId: camper.chargeId,
+          };
```
annechung, line 133: refactor here looks good, nice work!

annechung (approved):
tested and works well, awesome job with this 🚀

PR #23, opened 2022-03-05
feat: create waiver model and api endpoints

uyousafzai54:
## Notion ticket link

[Waiver Model + API](https://www.notion.so/uwblueprintexecs/3cc129328c8244d19e63596c79a430e3?v=54d9ce3263ca4749bb07bfb6e9a17803&p=c5eca2a9717d41e099d93a0937fb3601)

## Implementation description
* Created a new waiver model that contains an array of `Clause`'s. 

## Steps to test
1. Hit the `http://localhost:5000/admin/waiver` endpoint in Postman with a JSON object in the body containing an array of objects of the type `Clause`. An example body is shown below. 
2. The validator can be tested by sending an array of objects where one or more of the fields have an incorrect type. The server should respond with a status code `500` and the following error message `One or more objects in the array does not follow the schema of a Clause object.`

```
[
    {
        "text": "first",
        "required": true
    },
    {
        "text": "second",
        "required": false
    },
    {
        "text": "third",
        "required": false
    }
]
```

* Creation of a new Admin service. 

## Checklist


NirmalHegde (approved):

backend/typescript/models/waiver.model.ts
```diff
@@ -0,0 +1,39 @@
...
+    required: true,
+  },
+  e: {
+    type: String,
+    required: false,
+  },
+  f: {
+    type: String,
+    required: false,
+  },
```
annechung, line 36: hmm this works, but I think it might be better to change it so the number of clauses in the waiver is dynamic - we could store the clauses in a string array, for example. we just don't want to restrict the waiver to a strict number of fields in case that's subject to change. also, there should be some way of marking each clause as optional. let me know if you have any qs on this

backend/typescript/rest/campRoutes.ts
```diff
@@ -43,4 +44,30 @@ campRouter.get("/csv/:id", async (req, res) => {
   }
 });
 
+campRouter.post("/waiver", async (req, res) => {
```
annechung, line 47: also wanna add a validator here for checking input types

annechung (commented):
on second thought, can we actually make a separate service for updating admin form related things? we can call it FormService - thinking we could use this service for the form template get and edit implementations as well. so instead of `/camp/waiver` the endpoint would be `/forms/waiver`? let me know your thoughts on this

uyousafzai54:
> on second thought, can we actually make a separate service for updating admin form related things? we can call it AdminService - thinking we could use this service for form templates and promo codes going forwards as well. so instead of `/camp/waiver` the endpoint would be `/forms/waiver`? let me know your thoughts on this

Yep, I agree I also thought it was a bit weird putting the waiver endpoint under the camp service.

backend/typescript/rest/adminRoutes.ts
```diff
@@ -0,0 +1,35 @@
...
+
+
+
+adminRouter.post("/waiver", async (req, res) => {
+    try {
+      console.log(req.body)
+      const waiver = await adminService.updateWaiver({
+          clauses: req.body
+      })
+      res.status(200).json(waiver);
```
uyousafzai54: this can be changed to `waiver.clauses` if we only want to return the array of clauses

backend/typescript/services/implementations/adminService.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const Logger = logger(__filename);
+
+class AdminService implements IAdminService {
+    async updateWaiver(waiver: WaiverDTO): Promise<WaiverDTO> {
+        let waiverDto: WaiverDTO | null;
+        try{
+            let stuff = await waiverModel.updateOne(
+                {
+                  clauses: {'$exists': true}
```
uyousafzai54: This will create the clauses array if it doesn't exist.

backend/typescript/middlewares/validators/util.ts
```diff
@@ -34,6 +50,28 @@ export const validatePrimitive = (value: any, type: Type): boolean => {
   }
 };
 
+export const validateArrayOfObjects = (
```
annechung, line 53: hmm i feel like it might be easier to just have a function which validates a specific clause object (e.g. `validateWaiverClause`) and call that function on every clause in the waiver? the combination of this with `getType` seems a bit overcomplicated? and not very efficient since we need to getType for each key for every object. does anyone else have thoughts on thiss @uwblueprint/focus-on-nature

annechung (commented):
thanks for making the admin service - just a few comments on the waiver schema and the validation, we can go over this together later as well

backend/typescript/middlewares/validators/adminValidators.ts
```diff
@@ -0,0 +1,22 @@
...
+  const objArr: Array<any> = req.body;
+  objArr.forEach((obj) => {
+    if (obj.text && !validatePrimitive(obj.text, "string")) {
+      return res.status(400).send(getApiValidationError("text", "string"));
+    }
+    if (!validatePrimitive(obj.required, "boolean")) {
+      return res.status(400).send(getApiValidationError("required", "boolean"));
+    }
+    return null;
+  });
```
annechung, line 20: nit: might be a bit more readable if we created a helper function like validateClause and just called this function on each object, like
```
if (!req.body.clauses.every(validateClause) {
  return res.stats(400)...
}
```

annechung: can we also validate that req.body.clauses is non-empty? though not a huge deal since you're doing this in the updateWaiver function already

backend/typescript/models/waiver.model.ts
```diff
@@ -0,0 +1,26 @@
+import mongoose, { Schema, Document } from "mongoose";
+
+export interface Clause extends Document {
+  text: string;
+  required: boolean;
+}
+export interface Waiver extends Document {
+  clauses: Clause[];
+}
```
annechung, line 9: were you not able to nest the interface here like 

```
export interface Waiver extends Document {
  clauses: {
    text: string;
    required: boolean;
 }[]
}
```

backend/typescript/types.ts
```diff
@@ -100,3 +100,12 @@ export type NodemailerConfig = {
...
 };
+
+export type ClauseDTO = {
+  text: string;
+  required: boolean;
+};
+
+export type WaiverDTO = {
+  clauses: ClauseDTO[];
+};
```
annechung, line 111: same with nesting this?

backend/typescript/rest/adminRoutes.ts
```diff
@@ -0,0 +1,30 @@
...
+import IAdminService from "../services/interfaces/adminService";
+import { getErrorMessage } from "../utilities/errorUtils";
+import { waiverUpdateValidator } from "../middlewares/validators/adminValidators";
+
+const adminRouter: Router = Router();
+const adminService: IAdminService = new AdminService();
+adminRouter.post("/waiver", waiverUpdateValidator, async (req, res) => {
+  try {
+    const waiver = await adminService.updateWaiver({
+      clauses: req.body,
```
annechung, line 13: should this be req.body.clauses? it's causing an error on the backend for me if I send an object like 
```
{
    "clauses": [
        {
            "text": "hello",
            "required": true
        }
    ]
}
```

annechung (commented):
could you also delete the binary file that's attached to this PR? :0
few comments, should be ready to merge soon!

backend/typescript/middlewares/validators/adminValidators.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+  return true;
+};
+
+export const waiverUpdateValidator = async (
+  req: Request,
+  res: Response,
+  next: NextFunction,
+) => {
+  if (req.body.clauses.length === 0 || !req.body.clauses.every(validateClause)) {
```
annechung, line 21: DKM
```suggestion
  if (!req.body.clauses || req.body.clauses.length === 0 || !req.body.clauses.every(validateClause)) {
```

backend/typescript/middlewares/validators/adminValidators.ts
```diff
@@ -0,0 +1,29 @@
+import { Request, Response, NextFunction } from "express";
+import { validatePrimitive } from "./util";
+
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable-next-line import/prefer-default-export */
+const validateClause = (obj: any): boolean => {
+  if (obj.text && !validatePrimitive(obj.text, "string")) {
```
annechung, line 7:
```suggestion
  if (!validatePrimitive(obj.text, "string")) {
```

annechung (approved):
awesome workk :D

PR #30, opened 2022-03-12
feat: refund confirmation page API

NirmalHegde:
## Notion ticket link

[Ticket Name](https://uwblueprintexecs.notion.site/Refund-Confirmation-API-20bc2f90b63b4e6194ac8a00d8723730)

## Implementation description
* Implemented get refund confirmation to get info of user wanting to get refund.
* Provided direct link access for server

## Steps to test
1. ensure that 404, 500, and 200 result paths are reached
2. ensure right data comes back

* ensure 200 path provides correct info

## Checklist


backend/typescript/rest/camperRoutes.ts
```diff
@@ -76,4 +76,20 @@ camperRouter.get("/", async (req, res) => {
   }
 });
 
+camperRouter.get("/refund-confirm/:chargeId", async (req, res) => {
+  const { chargeId } = req.params;
+  try {
+    const camper = await camperService.getCamperByChargeId(
+      (chargeId as unknown) as number,
```
annechung, line 83: chargeId was updated to a string type

backend/typescript/services/implementations/camperService.ts
```diff
@@ -167,6 +167,41 @@ class CamperService implements ICamperService {
 
     return camperDtos;
   }
+
+  async getCamperByChargeId(chargeId: number): Promise<CamperDTO> {
```
annechung, line 171:
```suggestion
  async getCamperByChargeId(chargeId: string): Promise<CamperDTO> {
```

backend/typescript/services/interfaces/camperService.ts
```diff
@@ -23,6 +23,14 @@ interface ICamperService {
...
    */
   getCampersByCampId(campId: string): Promise<Array<CamperDTO>>;
+
+  /**
+   * Get campers associated with charge id
+   * @param chargeId camper's charge id for refunds
+   * @returns CamperDTO
+   * @throws Error if camper retrieval fails
+   */
+  getCamperByChargeId(chargeId: number): Promise<CamperDTO>;
```
annechung, line 33:
```suggestion
  getCamperByChargeId(chargeId: string): Promise<CamperDTO>;
```

backend/typescript/services/implementations/camperService.ts
```diff
@@ -167,6 +167,41 @@ class CamperService implements ICamperService {
...
+    try {
+      // eslint-disable-next-line prettier/prettier
+      const camper: Camper | null = await MgCamper.findOne({ "chargeId": chargeId });
```
annechung, line 174: there can be multiple campers with the same chargeId and we should be returning all of them

backend/typescript/services/implementations/camperService.ts
```diff
@@ -167,6 +167,41 @@ class CamperService implements ICamperService {
...
+
+      if (!camper) {
+        throw new Error(`Camper with Charge Id ${chargeId} not found.`);
+      }
+
+      const camperDTO: CamperDTO = {
```
annechung, line 180: should update this to use the newer CamperDTO

backend/typescript/rest/camperRoutes.ts
```diff
@@ -76,4 +76,20 @@ camperRouter.get("/", async (req, res) => {
...
+    );
+    res.status(200).json(camper);
+  } catch (error: unknown) {
+    if (getErrorMessage(error).includes("Camper with Charge Id")) {
```
annechung, line 87: this works but is a bit jank :'D would it be possible to use the cause property on the Error object before throwing it in the camper service and check that property here? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error

annechung (commented):
good start, but a few changes have to be made to return multiple campers and also to use the latest dto!

backend/typescript/rest/camperRoutes.ts
NirmalHegde: for sure!

annechung: returning just the one 500 error works tooo, we can leave it as it

annechung (approved):
tested and looks good!! can merge this in :D

PR #34, opened 2022-03-24
feat: add image upload

SinghHarman286:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Tasks-38b49f83f6ce4b07ae188dabb37631e8?p=1b2a2134dfa84cb1b461fea7a58a70a8)

## Implementation description
* added image upload middleware in `POST /camp` route, where image and camp data should be sent in `form-data`. `form-data` contains 2 keys: file containing the image that needs to be uploaded, and body containing camp data that needs to be saved. When image is successfully uploaded, a key: `fileName` containing image name gets stored in db along with the camp data
* "file" key in form-data is optional i.e. uploading an image in `POST /camp` route is optional. If the image is not provided, then `fileName` key will not be created in db
* max image upload size is 5MB

## Steps to test
1. In order to test the changes, go to postman->body->form-data. Create 2 keys: file containing the image that needs to be uploaded, and body containing the camp data that needs to be stored in db. You can use the below sample body input:
`{"active": true, "ageLower": 15, "ageUpper": 30, "name": "test-camp-image", "description": "image-test-description", "location": "canada", "capacity": 20,  "fee": 25,  "formQuestions": [{"question": "Q1", "type": "Text", "required": true, "description": "question description"}], "campSessions": []}`

Screenshot 2022-04-13 at 4 28 25 PM

In order to verify if correct image is uploaded, you can refer to this [link](https://console.firebase.google.com/u/6/project/focus-on-nature-6c0b1/storage/focus-on-nature-6c0b1.appspot.com/files)

* does `fileName` key needs to be created if there is no image to be uploaded?

## Checklist


amanda-guo (commented):
uploading an image works as expected. Also tested non-jpeg/png and a pic > 5 MB and this works as well (throws errors and no entries created in DB). Merge conflicts do need to be resolved to make sure the image upload is just for `camps`, so I can take a look at this again after that's done!

amanda-guo (commented):
One thing...not sure if this was in the scope of the ticket, but are there supposed to be changes to the edit camp API for the images or is that a separate ticket?

SinghHarman286:
> One thing...not sure if this was in the scope of the ticket, but are there supposed to be changes to the edit camp API for the images or is that a separate ticket?

Currently, there is no edit camp API and I believe a separate ticket needs to be made for that. cc @annechung

amanda-guo (approved):
works as expected! also i think we probably don't need to create a `fileName` key if there is no image uploaded and probably just add one if there is an image added later

SinghHarman286:
> works as expected! also i think we probably don't need to create a `fileName` key if there is no image uploaded and probably just add one if there is an image added later

I think that's the way it is happening currently. If no image is provided, then fileName key is not created: [link](https://github.com/uwblueprint/focus-on-nature/blob/00983dc48b40b52fe5a7010d2af2d43f75904f1e/backend/typescript/services/implementations/campService.ts#L181)

backend/typescript/middlewares/validators/campValidators.ts
```diff
@@ -92,11 +100,17 @@ export const createCampDtoValidator = async (
...
   }
-  if (req.body.campers) {
+  if (body.campers) {
     return res.status(400).send("campers should be empty");
   }
-  if (req.body.waitlist) {
+  if (body.waitlist) {
     return res.status(400).send("waitlist should be empty");
   }
+  if (req.file && !validateImageType(req.file.mimetype)) {
```
annechung, line 109: super nitpicky but: can we change the variable name to something a bit more descriptive (e.g. campPhoto?)

.gitignore
```diff
@@ -9,3 +9,4 @@
 .vscode
 **/*.cache
 **/*.egg-info
+**/uploads/
```
annechung, line 12: what does this directory store? :o

backend/typescript/rest/campRoutes.ts
```diff
@@ -19,25 +29,35 @@ campRouter.get("/", async (req, res) => {
...
+        location: body.location,
+        capacity: body.capacity,
+        fee: body.fee,
+        formQuestions: body.formQuestions,
+        campSessions: body.campSessions,
+        filePath: req.file?.path,
+        fileContentType: req.file?.mimetype,
+      });
+      if (req.file?.path) {
+        fs.unlinkSync(req.file.path);
```
annechung, line 53: what is this doing?

annechung (commented):
tested and the image upload works properly, nice work :) 
just confirming, the file name is the identifier for the image that's uploaded? (and we can use the file name to get the url of the image?)

backend/typescript/rest/campRoutes.ts
SinghHarman286: Each image thats uploaded is temporarily stored in "upload" directory, so this `fs.unlinkSync(req.file.path)` deletes the file from that directory. For example, req.file.path is something like 'uploads/152280faf4e59f64175f8a7186a48eab', so `fs.unlinkSync(req.file.path)` will just remove `152280faf4e59f64175f8a7186a48eab` from the upload directory

backend/typescript/middlewares/validators/campValidators.ts
SinghHarman286: oh you mean like something like validateCampPhotoType?

backend/typescript/middlewares/validators/campValidators.ts
```diff
@@ -15,39 +20,42 @@ export const createCampDtoValidator = async (
   res: Response,
   next: NextFunction,
 ) => {
-  if (!validatePrimitive(req.body.name, "string")) {
+  let body;
+  try {
+    body = JSON.parse(req.body.body);
```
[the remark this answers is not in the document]
SinghHarman286: I think formData can support either a string or a blob type. If we add any fields to formdata, then its value will be a string, so I guess we'll have write another logic to parse those value in their correct data type which I think might result in complex code and more errors? https://developer.mozilla.org/en-US/docs/Web/API/FormData/append

.gitignore
SinghHarman286: it stores the files that does not get uploaded because of invalid content type

SinghHarman286: As I am thinking a bit more about this, I think I should use `fs.unlinkSync` in the validator class as well just before we are returning invalid content type error response. That way, the files with invalid content type will not get stored in the uploads directory

backend/typescript/rest/campRoutes.ts
annechung: ah ok gotcha

backend/typescript/middlewares/validators/campValidators.ts
annechung: i meant `file`!

.gitignore
annechung: got it. removing invalid files sounds good to me, and yep should ignore uploads

backend/typescript/middlewares/validators/campValidators.ts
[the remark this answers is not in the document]
annechung: ah I see - we can keep it like this then! can we rename `body` to `data` though? `body.body` just looks funny to me. up to you, not a huge deal either way

backend/typescript/middlewares/validators/campValidators.ts
SinghHarman286: adding custom properties to `Request` object would result in typescript throwing an [error](https://stackoverflow.com/questions/58957228/property-does-not-exist-on-type-requestparamsdictionary). In order to fix it, we'll need to add additional changes to `Request`'s type definition which could make the code more complex. Thus, sticking with property name `file`.

SinghHarman286:
> tested and the image upload works properly, nice work :) just confirming, the file name is the identifier for the image that's uploaded? (and we can use the file name to get the url of the image?)

Yes, file name is the identifier and we'll use something like getFile(fileName: string) in order to retrieve the file -> https://uwblueprint.github.io/starter-code-v2/docs/file-storage/typescript-file-storage-service.html#getfile

annechung (approved):
looks good! I think there are cases where camp creation fails but the image will still be uploaded which we can look into separately, but this is definitely a strong start. awesome work :D

PR #41, opened 2022-04-02
feat: add capacity check on campsession update

rchang0501:
## Notion ticket link
[Registration API - capacity check enhancement](https://uwblueprintexecs.notion.site/Registration-API-capacity-check-enhancement-8ad08840bd72465a8a9b2732c18a3487)

## Implementation description
* moved camp capacity validation from camperService to camper.model
* this change was implemented to prevent race conditions from occurring upon registering a new camper to a camp session

## Steps to test
1. Make a new POST request at http://localhost:5000/camper/register
2. Set the campSession field to a camp that's full --> request should through an error stating "Error: camp is full"
3. Set the campSession field to a camp that's not full --> request should return request body as expected

* Are the error messages formatted correctly?
* Is there a more efficient way to do this?

## Checklist

### DIFF
--- a/backend/typescript/models/campSession.model.ts
+++ b/backend/typescript/models/campSession.model.ts
@@ -7,7 +7,7 @@ export interface CampSession extends Document {
   active: boolean;
   camp: Schema.Types.ObjectId;
   capacity: number;
-  campers: (Camper | Schema.Types.ObjectId)[];
+  campers: (Camper | Schema.Types.ObjectId | string)[];
   dates: Date[];
   endTime: string;
   priceId: string;
@@ -38,6 +38,12 @@ const CampSessionSchema: Schema = new Schema({
       },
     ],
     default: [],
+    validate: [
+      function (this: any, value: any) {
+        return value.length <= this.capacity;
+      },
+      "Camp is at maximum capacity",
+    ],
   },
   dates: {
     type: [Date],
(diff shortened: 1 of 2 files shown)


backend/typescript/services/implementations/camperService.ts
```diff
@@ -37,20 +37,21 @@ class CamperService implements ICamperService {
...
           if (!existingCampSession) {
             throw new Error(
               `Camp session ${campers[0].campSession} not found.`,
             );
           }
+
+          existingCampSession.campers = existingCampSession.campers
+            .map((id) => id.toString())
+            .concat(newCamperIds);
+          await existingCampSession.save();
```
annechung, line 53: ok so. we had to change from using findByIdAndUpdate to this because validators only work properly on document saving

annechung: for some reason converting from string to objectId was BROKEN. but I realized when you pass string values mongoose auto converts those strings to object ids. that's why I added string type to campers field in CampSession interface and I'm just passing strings.

backend/typescript/models/campSession.model.ts
```diff
@@ -38,6 +38,12 @@ const CampSessionSchema: Schema = new Schema({
       },
     ],
     default: [],
+    validate: [
+      function (this: any, value: any) {
+        return value.length <= this.capacity;
+      },
+      "Camp is at maximum capacity",
```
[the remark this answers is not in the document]
annechung: we talked offline but just to update here: we shouldn't have any auto waitlisting/registering logic so it's fine to just reject all campers and return an error message

PR #42, opened 2022-04-05
feat: bulk edit  and delete campers

amanda-guo:
## Notion ticket link

[Removing and moving campers in bulk](https://www.notion.so/uwblueprintexecs/Removing-and-moving-campers-in-bulk-e72e226660ca46688ce7ec3c0ca1a61e)

## Implementation description
* `updateCampersById` → takes in an array of `camperId`s and an `UpdateCamperDTO` that specifies the fields that need to be updated for all the campers specified (via the `camperId`s). Update happens as long as all campers have the same `chargeId`. If `campSession` is passed as a field to be updated, then the `campers` array in both the new and old `campSession`s are updated.
* Note: if want to update distinct fields for different campers, will have to call this update function for each camper/fields to be updated
* `deleteCampersById` → can take in an arbitrary number of `camperId`s, and as long as they are all existing campers with the same `chargeId` then they can be deleted
* `cancelRegistration` → can take in an arbitrary number of `camperId`s and a `chargeId` associated with all the campers, then does a check for whether the camp session that these campers are registered for starts more than 30 days after the cancellation OR if the waitlist for the camp is not empty and is less than 30 days from cancellation to start date. Then calls the `deleteCampersById` function if these checks pass
* Assumption for all these functions: `chargeId` is the parameter that establishes that these campers are all from the same family (can be edited/delete together) - this may need to change later

## Steps to test `updateCampersById`
1. Use this testing endpoint on Postman: http://localhost:5000/campers/
2. Pass in `req.body`: an array of `camperId`s under `camperIds` and the camper fields that need to be updated (e.g. {"camperIds": [...], "campSession": ...}). Format like this
```
{
    "camperIds": [
        "625786928d4e72b0e41193a6",
        "62576da892c3a430022c653d"
    ],
    "firstName": "JANET",
    // add any other fields to be updated

}
```
3. Test cases: updating one or many campers (making sure the same fields are updated), omitting certain fields in the update DTO, testing that the update still works and testing that the `camper` array in the`campSession`s are also updated if that field is passed
4. Error cases (these cases should throw errors): moving campers between `campSession`s that are not under the same camp, wrong primitive values passed, `camperId`s for campers that don’t exist, `campSession` that doesn’t exist, different `chargeId`s on each camper, invalid `campSession`

## Steps to test `deleteCampersById`
1. Use this testing endpoint on Postman: http://localhost:5000/campers/
2. Pass in `req.body`: an array of `camperId`s (e.g. "camperIds" : ["`6248a739dd79f2313c922c85`", "`6248a73add79f2313c922c8a`"])
3. Test cases: deleting 0, 1, 2+ campers, checking that the campers are removed from the `campSession`'s array of `campers`
4. Error cases (these cases should throw errors): campers who don’t have the same `chargeId`, campers that don’t exist in the DB

## Steps to test `cancelRegistration`
1. Use this testing endpoint on Postman: http://localhost:5000/campers/cancel-registration
2. Pass in `req.body`: an array of `camperId`s (e.g. "camperIds" : ["`6248a739dd79f2313c922c85`", "`6248a73add79f2313c922c8a`"]) and a `chargeId`
3. Test cases: cancelling registration for 0, 1, 2+ campers, cancelling registration if the `campSession` start date > 30 days or waitlist is not empty, check that refund is made on stripe and that campers are removed from the `campSession`'s array of `campers`
4. Error cases (these cases should throw errors): campers that are not all registered for the same `campSession`, campers that don’t exist in the DB, camp session start date < 30 days and waitlist is empty

* Does the code make sense for the use cases it is made for?

## Checklist

### DIFF
--- a/backend/typescript/services/implementations/camperService.ts
+++ b/backend/typescript/services/implementations/camperService.ts
@@ -1,3 +1,4 @@
+import Stripe from "stripe";
 import ICamperService from "../interfaces/camperService";
 import MgCamper, { Camper } from "../../models/camper.model";
 import MgWaitlistedCamper, {
@@ -5,7 +6,7 @@ import MgWaitlistedCamper, {
 } from "../../models/waitlistedCamper.model";
 import MgCampSession, { CampSession } from "../../models/campSession.model";
 import {
-  CreateCamperDTO,
+  CreateCampersDTO,
   CamperDTO,
   CreateWaitlistedCamperDTO,
   WaitlistedCamperDTO,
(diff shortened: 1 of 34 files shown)


backend/typescript/services/interfaces/camperService.ts
```diff
@@ -54,30 +54,30 @@ interface ICamperService {
...
 
   /**
-   * Delete all campers associated with the charge ID
-   * @param chargeId the charge ID for the payment
+   * Delete all campers associated with the camper IDs if the camp session start date is > 30 days from this cancellation request OR the waitlist for that camp session is not empty and the camp session start date is <= 30 days from this cancellation request
+   * @param camperIds array of camper Ids
    * @throws Error if camper cancellation fails
    */
-  deleteCampersByChargeId(chargeId: string): void;
+  cancelRegistration(camperIds: Array<string>): void;
```
jjjding2, line 73: return type should be `Promise<void>`

backend/typescript/services/interfaces/camperService.ts
```diff
@@ -54,30 +54,30 @@ interface ICamperService {
...
   /**
-   * Delete camper associated with the camper ID
-   * @param camperId camper's Id
-   * @throws Error if camper cancellation fails
+   * Delete campers associated with the camper IDs
+   * @param camperIds array of camper Ids
+   * @throws Error if camper deletion fails
    */
-  deleteCamperById(camperId: string): void;
+  deleteCampersById(camperIds: Array<string>): void;
```
jjjding2, line 80: Same here: `deleteCampersById(camperIds: Array<string>): Promise<void>;`

backend/typescript/services/implementations/camperService.ts
```diff
@@ -571,56 +687,114 @@ class CamperService implements ICamperService {
...
+          },
+        });
+        /* eslint-disable no-await-in-loop */
+        for (let i = 0; i < campers.length; i += 1) {
+          await emailService.sendAdminCamperCancellationNoticeEmail(
+            camp,
+            campers[i],
+            campSession,
+          );
+        }
```
annechung, line 787: need to move to cancelRegistration method

backend/typescript/services/implementations/camperService.ts
```diff
@@ -571,56 +687,114 @@ class CamperService implements ICamperService {
...
+          throw new Error(
+            `Campers' camp session with campSessionId ${campers[i].campSession} not found.`,
+          );
+        }
+        if (camperCampSession.id.toString() !== campSession.id.toString()) {
+          throw new Error(
+            `Not all campers are registered for the same campSession.`,
+          );
+        }
+      }
```
annechung, line 735: like my other comment above we have `campSessionId` in campers so we can do this check without retrieving each campSession

AmandaMisjuwar (approved):

PR #47, opened 2022-04-16
feat: add camp registration endpoints

jjjding2:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Tasks-38b49f83f6ce4b07ae188dabb37631e8?p=d436ed7850ba48f28d8b3fd09cd66015)

* Leaving formQuestions to another PR

## Implementation description
* Added endpoints for updating Camp, and creating/updating/deleting a CampSession
   * PATCH /camp/
   * POST/PATCH/DELETE /camp/:campId/session
* Went through registration designs to figure out which APIs should be called and when
* All CampSession creation/deletion should be done through the /session endpoints, rather than modifying /camp/ directly
* `campers` and `waitlist` fields are never modified by these endpoints, to avoid payment issues

## Steps to test
1. Verify automated tests: (`docker exec -it scv2_ts_backend /bin/bash -c "yarn test"`)
2. Try updating Camp by calling `localhost:5000/camp/:campId`
3. Try adding CampSession to Camp with `localhost:5000/camp/:campId/session/`
4. Try updating CampSession in Camp with `localhost:5000/camp/:campId/session/:campSessionId`
5. Try deleting CampSession in Camp with `localhost:5000/camp/:campId/session/:campSessionId`

Test Camp Body:
```
{
    "ageLower": 15,
    "ageUpper": 30,
    "name": "test camp",
    "description": "description",
    "location": "canada",
    "fee": 25
}
```
Test CampSession Body:
```
{
    "campSessions": [
        {
            "capacity": 20,
            "active": true,
            "startTime": "12:00",
            "endTime": "17:00",
            "dates": [
                "Sun Mar 13 2022 20:01:14 GMT-0600 (Mountain Daylight Time)"
            ]
        }
    ]
}
```
Test FormQuestion Body: 
```
{
"question": "Q1",
"type": "Multiselect",
"required": true,
"description": "question description"
}
```

* Verify that the endpoints catch whether each CampSession is contained in the associated `Camp` (i.e. you must pass in the id of the Camp containing the session you want to delete)
* Verify that no functionality is missing for the registration flow

## Checklist

TODO
- [] be able to support bulk campSession creation, sort dates, check for overlap


backend/typescript/types.ts
```diff
@@ -120,6 +120,21 @@ export type CreateCampDTO = Omit<
...
 };
 
+export type UpdateCampDTO = Omit<
+  CampDTO,
+  "id" | "formQuestions" | "campSessions"
+>;
+
+export type CreateCampSessionDTO = Omit<
+  CampSessionDTO,
+  "id" | "camp" | "campers" | "waitlist"
```
jjjding2, line 130: We currently support passing in CampSessions and FormQuestions to construct them at once – but in practice only basic details like description/fee will be called, and CampSessions / FormQuestions will be added one by one in a later section of registration

annechung (commented):
initial comments - have a few questions about how this works so will ask offline!

backend/typescript/services/implementations/campService.ts
```diff
@@ -85,6 +92,201 @@ class CampService implements ICampService {
     }
   }
 
+  async updateCampById(campId: string, camp: UpdateCampDTO): Promise<CampDTO> {
```
annechung, line 95: we also have to handle updating the camp photo if that's changed - will create a separate task for this

backend/typescript/services/implementations/campService.ts
```diff
@@ -85,6 +92,201 @@ class CampService implements ICampService {
...
+        !oldSessions.find((session) => session.toString() === campSessionId)
+      ) {
+        throw new Error(
+          `CampSession with campSessionId ${campSessionId} not found for Camp with id ${campId}.`,
+        );
+      }
+
+      const newCampSession: CampSession | null = await MgCampSession.findByIdAndUpdate(
+        campSessionId,
+        {
```
annechung, line 222: noticed we can't update capacity, but I think it should be editable as long as the camp hasn't been published. we can make a new task also for updating logic for when we can/can't update fields like fee and capacity on camps/campsessions

annechung (approved):
awesome work - single handedly took on all of the camps endpoints :0 we can create followup tasks for the formquestion endpoints, camp photo update on camp edit and status/other checks on camp edits but the bulk of the work is done now :D great job!!

backend/typescript/services/implementations/campService.ts
jjjding2: oops yeah we should def be able to update it here, must have missed this when capacity was moved to session. sounds good!

PR #49, opened 2022-04-21
Add productId and priceId to Camp model

jjjding2:
Small model change

## Checklist

### DIFF
--- a/backend/typescript/models/camp.model.ts
+++ b/backend/typescript/models/camp.model.ts
@@ -7,13 +7,14 @@ export interface Camp extends Document {
   ageLower: number;
   ageUpper: number;
   capacity: number;
-  name: string;
-  description: string;
-  location: string;
-  fee: number;
-  formQuestions: (Schema.Types.ObjectId | FormQuestion)[];
   campSessions: (Schema.Types.ObjectId | CampSession)[];
+  description: string;
+  fee: number;
   fileName?: string;
+  formQuestions: (Schema.Types.ObjectId | FormQuestion)[];
+  location: string;
+  name: string;
+  productId: string;
 }
 
 const CampSchema: Schema = new Schema({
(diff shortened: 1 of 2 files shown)


annechung (approved):

backend/typescript/models/campSession.model.ts
```diff
@@ -28,31 +33,30 @@ const CampSessionSchema: Schema = new Schema({
...
     type: [Date],
     required: true,
   },
-  startTime: {
+  endTime: {
     type: String,
     required: true,
   },
-  endTime: {
+  priceId: {
```
SinghHarman286, line 44: I was thinking that in `priceId`, we have `required` as false and have `default` as "". The reason being is that price Ids will be created only once `campSessions` are successfully created. So having them as required, we'll need to pass an empty string every time we are creating a campSession

backend/typescript/models/camp.model.ts
```diff
@@ -7,13 +7,14 @@ export interface Camp extends Document {
...
   description: string;
-  location: string;
   fee: number;
-  formQuestions: (Schema.Types.ObjectId | FormQuestion)[];
-  campSessions: (Schema.Types.ObjectId | CampSession)[];
   fileName?: string;
+  formQuestions: (Schema.Types.ObjectId | FormQuestion)[];
+  location: string;
+  name: string;
+  productId: string;
```
SinghHarman286, line 17: Do we need productId in camp model if we are following `prod_<mongoose_id>` syntax for stripe products?

SinghHarman286 (commented):
Looks good! Just left some comments/questions I have about some of the changes

backend/typescript/models/camp.model.ts
jjjding2: Decided to let Stripe generate the id stored in productId

PR #69, opened 2022-06-08
feat: update edit camp endpoint to accept file for thumbnail photo

uyousafzai54:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Camp-Edit-Thumbnail-911c5d6d93484863b4f98efa65ecb52a)

## Implementation description
* Updated the edit camp endpoint to accept a file and upload it to firebase for admins to edit the thumbnail photo of a camp.

## Steps to test
1.

* 

## Checklist


annechung (changes requested):
we need to update the validator to check request bodies of type form-data now, check the route/validator for create camp!

backend/typescript/rest/campRoutes.ts
```diff
@@ -82,30 +82,39 @@ campRouter.post(
...
-    });
-
-    res.status(200).json(newCamp);
-  } catch (error: unknown) {
-    res.status(500).json({ error: getErrorMessage(error) });
-  }
-});
+campRouter.patch(
+  "/:campId",
+  upload.single("file"),
```
annechung, line 87: can we delete the existing file here if there is one? but we should only delete if a new file is passed

backend/typescript/rest/campRoutes.ts
```diff
@@ -82,30 +82,39 @@ campRouter.post(
...
+  updateCampDtoValidator,
+  async (req, res) => {
+    try {
```
annechung, line 90: the camp data should be in req.body.data instead of req.body since we're changing the type of the request body to be form-data instead of json to be able to attach files, take a look at the route for creating camps: https://github.com/uwblueprint/focus-on-nature/blob/a8bf52a03438b19ad5921155939f7f8c6d934ace/backend/typescript/rest/campRoutes.ts#L52 let's also update the validator

annechung (changes requested):
almost ready! just a few fixes to make

backend/typescript/rest/campRoutes.ts
uyousafzai54: does it matter if we delete and reupload the same file?

annechung: hmm I think this is just checking the file path, not the file contents

backend/typescript/rest/campRoutes.ts
```diff
@@ -94,30 +94,48 @@ campRouter.post(
...
+        startTime: "",
+        volunteers: [],
+      });
+      if (req.file?.path) {
+        fs.unlinkSync(req.file.path);
+      } else {
+        const oldPath = oldCamp.filePath;
+        var oldImage = fs.readFileSync(oldPath);
+        var newImage = fs.readFileSync(req.file?.path);
+        if(!oldImage.equals(newImage)) {
```
annechung, line 129: hmm I think this is just checking the file path, not the file contents

annechung (approved):

PR #72, opened 2022-06-18
feat: get camp by camp id

rchang0501:
## Notion ticket link

[Get Camp by camp id](https://www.notion.so/uwblueprintexecs/Get-Camp-by-camp-id-a2d958a9e0de4f60bc3c47dd5a5968db)

## Implementation description
* added a getCampById end point to retrieve a camp by its Id 

## Steps to test
1. send a get request at `http://localhost:5000/camp/:id` 

* does the endpoint return all requested information? 

## Checklist

### DIFF
--- a/backend/typescript/rest/campRoutes.ts
+++ b/backend/typescript/rest/campRoutes.ts
@@ -27,6 +27,12 @@ const campService: ICampService = new CampService(fileStorageService);
 campRouter.get("/", async (req, res) => {
   try {
     const camps = await campService.getCamps();
+
+    if (req.params.id) {
+      await campService.getCampById(req.params.id);
+      res.status(204).send();
+    }
+
     res.status(200).json(camps);
   } catch (error: unknown) {
     res.status(500).json({ error: getErrorMessage(error) });
--- a/backend/typescript/rest/campRoutes.ts
+++ b/backend/typescript/rest/campRoutes.ts
@@ -12,6 +12,7 @@ import {
   updateCampDtoValidator,
   updateCampSessionDtoValidator,
 } from "../middlewares/validators/campValidators";
+import camperRouter from "./camperRoutes";
 
 const upload = multer({ dest: "uploads/" });
 
@@ -34,6 +35,16 @@ campRouter.get("/", async (req, res) => {
     }
 
     res.status(200).json(camps);
+  } catch (error: unknown) {
+    res.status(500).json({ error: getErrorMessage(error) });
+  }
+});
+
+campRouter.get("/:id", async (req, res) => {
+  try {
+    const camp = await campService.getCampById(req.params.id);
+
+    res.status(200).json(camp);
   } catch (error: unknown) {
     res.status(500).json({ error: getErrorMessage(error) });
   }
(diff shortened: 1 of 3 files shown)


backend/typescript/rest/campRoutes.ts
```diff
@@ -33,6 +33,16 @@ campRouter.get("/", async (req, res) => {
   }
 });
 
+campRouter.get("/:id", async (req, res) => {
+  try {
+    const camp = await campService.getCampById(req.params.id);
+
+    res.status(200).json(camp);
+  } catch (error: unknown) {
+    res.status(500).json({ error: getErrorMessage(error) });
```
NirmalHegde, line 42: If the user sends a get request with a wrong camp id, it should return a 400 series error, not 500. recommend adding an if statement or something to verify which error is being sent back.

NirmalHegde: This can also be done using some kind of validator middleware maybe?

NirmalHegde (commented):
Everything LGTM and works as expected! Just left a smol comment

backend/typescript/services/implementations/campService.ts
```diff
@@ -101,6 +101,47 @@ class CampService implements ICampService {
     }
   }
 
+  async getCampById(campId: string): Promise<CampDTO> {
+    let camp: Camp | null;
+
+    try {
+      camp = await MgCamp.findById(campId);
```
annechung, line 108: can we populate the camp sessions and form questions here so we return the objects instead of just the ids? can follow the example in getCamps https://github.com/uwblueprint/focus-on-nature/blob/de6fbd0d12001f6c633eb01b69b5a728b49389d0/backend/typescript/services/implementations/campService.ts#L34-L44

backend/typescript/services/interfaces/campService.ts
```diff
@@ -18,6 +18,14 @@ interface ICampService {
    */
   getCamps(): Promise<GetCampDTO[]>;
 
+  /**
+   * Get camp with the specified campId
+   * @param campId camp's id
+   * @returns a CampDTO
+   * @throws Error if camper retrieval fails
+   */
+  getCampById(campId: string): Promise<CampDTO>;
```
annechung, line 27: update return type here since we're populating the questions and sessions fields!
```suggestion
  getCampById(campId: string): Promise<GetCampDTO>;
```

annechung (changes requested):
good start - just need a change to return the form questions and sessions with the camp since we'll be using this to show the camp details pages.

backend/typescript/services/implementations/campService.ts
```diff
@@ -101,6 +101,47 @@ class CampService implements ICampService {
...
+
+      if (!camp) {
+        throw new Error(`Camp' with campId ${campId} not found.`);
+      }
+    } catch (error: unknown) {
+      Logger.error(`Failed to update camp. Reason = ${getErrorMessage(error)}`);
```
annechung, line 114: update error message!
```suggestion
      Logger.error(`Failed to get camp. Reason = ${getErrorMessage(error)}`);
```

backend/typescript/rest/campRoutes.ts
annechung: I think this is fine to leave as is! returning the error message is good enough imo, we can display the message in the frontend

annechung (approved):
looks good! 🚀

PR #76, opened 2022-07-02
Feat: create stripe products/prices

jjjding2:
## Notion ticket link

[Creating Products - Stripe API](https://www.notion.so/uwblueprintexecs/Task-Board-S22-1a3e651ab8544192bf0fdcbba04f6d03?v=d58855b8ca38450babd8d19c3665919b&p=e604a09bb2ef4c69b587a1ffbf3d4549)

## Implementation description
* Add products/prices for camps, latePickUp, and earlyDropOff
* Removed functionality to create campSessions as part of createCamp (must call createCampSessions endpoint)

_Model_ - Fees
* ~~Add new model to store latePickUp and earlyDropOff fees which are fetched from db (currently both fixed as 5)~~
* Pass fees as param

_Route_ - Create Camp
* Create Stripe products on camp creation using StripeUtil functions
* Prepend camp names to earlyDropOff / latePickUp products

_Route_ - Update Camp
* If Stripe products are missing, create them
* Update the name/description of existing Products with updateStripeCampProduct
* Create **Price** objects if the Camp is switching from inactive to active
* Prevent any modification to fields that would change the price after camp has been set to active (i..e increasing # of dates)

_Route_ - Create Camp Sessions
* If Camp is active, create Price objects for existing campSessions
* Otherwise, let their priceIds be empty strings

## Steps to test
**[Sample JSON bodies](https://www.notion.so/uwblueprintexecs/Sample-JSON-Bodies-1598b0c3ebbc4db9b46a2b735222b9f7)**
1. ``POST localhost:5000/camp`` Create an inactive camp (Verify Stripe products are created)
```javascript
{
	"ageLower": 15,
	"ageUpper": 30,
	"campCoordinators": ["61fb3d34272ea0002ad6a24d"],
	"campCounsellors": ["61fb3d34272ea0002ad6a24d"],
	"name": "test camp",
	"description": "description",
	"earlyDropoff": "12:30",
	"latePickup": "2:30",
	"location": {
            "streetAddress1": "123 Focus on Nature Avenue",
            "city": "Guelph",
            "province": "Ontario",
            "postalCode": "M1A 3G3"
         },
	"fee": 25,
	"volunteers": ["jason"],
	"startTime":"12:30",
	"endTime":"2:30",
	"active":false
}
```
2. ``POST localhost:5000/camp/:campID/session/`` Add Camp Sessions (prices should not be created)
```javascript
"campSessions": [
        {
            "capacity": 20,
            "dates": [
                "Sun Mar 13 2022 20:01:14 GMT-0600 (Mountain Daylight Time)"
            ]
        }
]
```
3. ``PATCH localhost:5000/camp/:campID`` Update Camp to active - verify prices are created

* Verifying that Price/Product objects are created at the right time
* Verifying that fields cannot be improperly changed after a camp has been set to active
* Does this work with partial/incomplete camps in the collection?

## Checklist


jjjding2:
Need to pass in fees as a parameter instead of using a constant in the database

annechung (changes requested):
this is really great jason! tested it and it works well. I think it makes more sense though to share dropoff and pickup price objects across camp sessions in the same camp since the amounts don't differ across sessions. can we create one price object for dropoff and pickup each when active is set to true on the camp and set dropoffPriceId and pickupPriceId fields on the camp model?

PR #79, opened 2022-07-09
Feat: Restrict Access to Registration Form

NirmalHegde:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Waitlist-Restrict-access-to-registration-form-caf825d88fac4a83a04791c03914ea43)

## Implementation description
* Added restrictions for waitlisted applicants on the getCampById endpoint such that users cannot sign up for a camp when not available
* Added query parameters for waitlisted campers and camp session id and added checks to ensure that the link clicked is not expired

## Steps to test
1. Trigger endpoint /camp/{id}?wId={waitlistedCamperId}&session={campSessionId} and see if results returned are expected

* Ensure edge cases are accounted for

## Checklist


backend/typescript/services/implementations/campService.ts
```diff
@@ -125,6 +160,18 @@ class CampService implements ICampService {
...
+        (campSession) => campSession.id === campSessionId,
+      );
+
+      if (camp.campSessions.length === 0) {
+        throw new Error(
+          `No camp session matches with session ${campSessionId}`,
+        );
+      }
+    }
+
```
annechung, line 174: can we also check the camp session's capacity here and throw an error if it's full? I think it's better than doing the check on the frontend, avoids doing a redirect? what do you think

NirmalHegde: I think we agreed that the check would be done on the frontend, but yeah I can add that.

backend/typescript/services/implementations/campService.ts
```diff
@@ -103,9 +106,41 @@ class CampService implements ICampService {
...
+      try {
+        const waitlistedCamper: WaitlistedCamper | null = await MgWaitlistedCamper.findById(
+          waitlistedCamperId,
+        );
+
+        if (!waitlistedCamper || !waitlistedCamper?.linkExpiry) {
+          throw new Error(
+            `Waitlisted Camper with Id ${waitlistedCamperId} does not exist or does not have an invite link.`,
+          );
+        }
```
annechung, line 125: can we check here that `campSessionId == waitlistedCamper.campSession`?

backend/typescript/services/implementations/campService.ts
```diff
@@ -103,9 +106,47 @@ class CampService implements ICampService {
...
 
+        if (campSessionId !== waitlistedCamper.campSession.toString()) {
+          throw new Error(
+            `Camp session ID ${campSessionId} does not exist or does not have an invite link.`,
```
annechung, line 128: nit: better error msg?

```suggestion
            `Given waitlisted camper is not on the waitlist for given camp session.`,
```

backend/typescript/services/implementations/campService.ts
```diff
@@ -103,9 +106,47 @@ class CampService implements ICampService {
...
 
+        if (campSessionId !== waitlistedCamper.campSession.toString()) {
+          throw new Error(
+            `Given waitlisted camper is not on the waitlist for given camp session.`,
+          );
+        }
+
+        const linkExpiryDate = new Date(waitlistedCamper.linkExpiry);
+        const currentDate = new Date();
+        if (linkExpiryDate > currentDate) {
```
annechung, line 134: should be <=? 
```suggestion
        if (linkExpiryDate <= currentDate) {
```

NirmalHegde: I think if they are registering on the current date, it should be allowed, but good catch

annechung (approved):
lgtm :D

PR #81, opened 2022-07-14
Feat: Update Waitlisted Camper to "Registered" after Camper is Created

NirmalHegde:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Waitlist-Update-status-of-waitlisted-camper-who-registers-4e33b329a79f41749509158c745e39ec)

## Implementation description
* Added check to ensure only one camper is being updated and then updated the field on waitlisted camper

## Steps to test
1. ensure endpoint works as expected and changes a waitlisted camper put into query params

* Right now, we are making createCampers work as an endpoint that returns multiple campers. However, we can only change one waitlisted camper at a time since there is no other way to relate a waitlisted camper to their information without it getting unnecessarily complex. As such, should this be abstracted out into another endpoint that would only create one camper (would call createCampers function but only feed one camper) and all the waitlistedCamper logic would go into that?

## Checklist

### DIFF
--- a/backend/typescript/services/implementations/camperService.ts
+++ b/backend/typescript/services/implementations/camperService.ts
@@ -147,13 +147,22 @@ class CamperService implements ICamperService {
     });
 
     if (waitlistedCamperId) {
-      await MgWaitlistedCamper.findByIdAndUpdate(
-        waitlistedCamperId,
-        {
-          status: "Registered",
-        },
-        { runValidators: true },
-      );
+      try {
+        await MgWaitlistedCamper.findByIdAndUpdate(
+          waitlistedCamperId,
+          {
+            status: "Registered",
+          },
+          { runValidators: true },
+        );
+      } catch (error) {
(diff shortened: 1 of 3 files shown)


backend/typescript/services/implementations/camperService.ts
```diff
@@ -136,6 +145,17 @@ class CamperService implements ICamperService {
         optionalClauses: newCamper.optionalClauses,
       };
     });
+
+    if (waitlistedCamperId) {
```
annechung, line 149: can we add a try/catch here so if the waitlistedCamperId is invalid we return an error message like `` `Registered campers but was unable to update waitlisted camper status with id: ${waitlistedCamperId}` ``?

annechung (changes requested):
looks good! just one smol comment

annechung (approved):
nice work :DD

PR #82, opened 2022-07-16
feat: camp overview information

amanda-guo:
## Notion ticket link

[Camp Overview: Camp level information view only](https://www.notion.so/uwblueprintexecs/Camp-Overview-Camp-level-information-view-only-343f864f624945edacc3d6fe599c0ebf)

## Implementation description
* Created a camp overview page that just includes the top-level camp information according to the Figma designs
* Added an APIClient to call the BE APIs to get the camp information
* There are some aspects of this overview that are out of scope of this ticket and some things that are blocked too (check Additional Notes & Clarifications in the Notion ticket to see what's included and not included)
* Currently a camp is just hard coded for now
* Some "TODOs" have been called out as comments in the code (left those there as placeholders for future updates that were deemed out-of-scope for this ticket)

## Steps to test
1. Build containers and open the page at http://localhost:3000/admin/camp/62c098e7b4a7a433a7622ff4 on a browser
2. Check that this matches the Figma designs and that the right information is pulled from the DB

* Code quality and placement

## Checklist


frontend/src/App.tsx
```diff
@@ -102,6 +103,15 @@ const App = (): React.ReactElement => {
                   path={Routes.HOOKS_PAGE}
                   component={HooksDemo}
                 />
+                {/* TODO: change this so it's not hardcoded */}
+                <PrivateRoute
+                  exact
+                  path={Routes.CAMP_OVERVIEW_PAGE.replace(
```
dhmi19, line 109: Why do we need to replace the ID parameter here? I tried removing the .replace() function and then reloading the page with the given url and it still worked. Removing this here would save us the refactor in the future maybe?

code:
`<PrivateRoute
                  exact
                  path={Routes.CAMP_OVERVIEW_PAGE}
                  component={CampOverview}
                />`

URL tested: http://localhost:3000/admin/camp/62c098e7b4a7a433a7622ff4

frontend/src/constants/AuthConstants.ts
```diff
@@ -1,3 +1,10 @@
+import { getLocalStorageObjProperty } from "../utils/LocalStorageUtils";
+
 const AUTHENTICATED_USER_KEY = `${window.location.hostname}:AUTHENTICATED_USER`;
 
+export const BEARER_TOKEN = `Bearer ${getLocalStorageObjProperty(
```
dhmi19, line 5: Was this part of another ticket? If so, just wanted to make sure this is not gonna create any conflicts on merge?

frontend/src/types/CampsTypes.ts
```diff
@@ -0,0 +1,40 @@
+export type Camp = {
+  id: string;
+  active: boolean;
+  ageLower: number;
+  ageUpper: number;
+  campCoordinators: string[];
+  campCounsellors: string[];
+  name: string;
+  description: string;
+  earlyDropoff: string;
```
dhmi19, line 10: Is it ok for these fields to be strings? In the future, we'll need to create 30 minute timeslots from Early Dropoff time to start time AND endTime to latePickup time. Would storing this as DateTime (with a constant Date) make things easier in the future? CC: @annechung @AmandaMisjuwar

dhmi19 (changes requested):
Looks good to me overall. Just left a few minor comments. I think there will be a few design changes needed as designers need to fix the Frame width and adjust the designs accordingly (everything looks really big right now). We also forgot to display the EDLP timings on the camp level info so we'll need to add that too (although, I think we can merge this PR if it gets blocked by those design changes)

frontend/src/constants/AuthConstants.ts
AmandaMisjuwar: We need this line for other tickets as well so it'll be fine to leave

dhmi19: sounds good!

frontend/src/components/pages/CampOverview/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+export type CampOverviewProps = {
+  campId: string;
+};
+
+const CampOverview = (): JSX.Element => {
+  const [camp, setCamp] = useState<Camp>();
+  const campId = "62c098e7b4a7a433a7622ff4"; // hardcoded for now, TODO: update this
+
+  // TODO: update the statuses
+  enum Status {
```
AmandaMisjuwar, line 29: Can we call this `CampStatus` instead so not to confuse it with `UserStatus` in otherr components?

frontend/src/types/CampsTypes.ts
```diff
@@ -0,0 +1,40 @@
+export type Camp = {
```
AmandaMisjuwar, line 1:
```suggestion
export type CampResponse = {
```

frontend/src/APIClients/CampsAPIClient.ts
```diff
@@ -0,0 +1,18 @@
...
+import baseAPIClient from "./BaseAPIClient";
+
+const getCampById = async (id: string): Promise<Camp> => {
+  try {
+    const { data } = await baseAPIClient.get(`/camp/${id}`, {
+      headers: { Authorization: BEARER_TOKEN },
+    });
+    return data;
+  } catch (error) {
+    return error as Camp;
```
AmandaMisjuwar, line 12: Can we actually call this CampResponse instead so we can stay consistent with our types for all the backend responses/return values?

```suggestion
    return error as CampResponse;
```

frontend/src/components/pages/CampOverview/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+    PUBLISHED = "Published",
+    DRAFT = "Draft",
+  }
+
+  const status = camp?.active ? Status.PUBLISHED : Status.DRAFT;
+
+  useEffect(() => {
+    const getCampInfo = async () => {
+      const campResponse = await CampsAPIClient.getCampById(campId);
+      setCamp(campResponse);
```
AmandaMisjuwar, line 39: We should have some kind of error handling for if campResponse is `null`
```suggestion
      if (campResponse) setCamp(campResponse);
      else // display an error or just let it remain null
```

frontend/src/components/pages/CampOverview/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+  }, []);
+
+  return (
+    <>
+      <Container maxWidth="100vw">
+        <Flex marginLeft="80px" marginRight="80px">
+          <Box width="100%" mt="1rem">
+            <HStack width="100%" marginBottom="8px" alignItems="center">
+              <Text align="left" textStyle="displayXLarge" marginBottom="8px">
+                {camp?.name}
```
AmandaMisjuwar, line 51: related to the previous comment about handling null campResponse/errors, we won't need the ? potentially

frontend/src/types/CampsTypes.ts
AmandaMisjuwar: Yea if we need to do date/time calculations we should store them as Dates (however in Mongo they're currently also strings). But if they're always 30 min time slots we could also just convert from string -> Date -> string whenever we do the calculation? @annechung @dhmi19

AmandaMisjuwar (changes requested):

frontend/src/App.tsx
annechung: yeah we can remove this

frontend/src/types/CampsTypes.ts
annechung: I feel like it makes more sense to store as strings since these fields are actually times that aren’t tied to a specific date?

AmandaMisjuwar: Yeah we'll just have to make sure theyre in a format that can be converted into Date if needed for calculation (will leave it as string for now)

annechung (approved):
cannot wait to overview all the camps

AmandaMisjuwar (approved):

PR #88, opened 2022-08-23
feat: camp overview registration list table + tabs

rchang0501:
## Notion ticket link

[Camp Overview: Registration List Table (View Only) + Tabs](https://www.notion.so/uwblueprintexecs/Camp-Overview-Registration-List-Table-View-Only-Tabs-88c011a405ce452fb573113c0fa4f8e4)

## Implementation description
* created tabs to switch between registration list and waitlist table
* created registration list table 

Screen Shot 2022-08-25 at 12 13 54 AM

## Steps to test
1. Go to the camps page to view the table
2. In `frontend/src/components/pages/camps/CampersTableNavBar.tsx` there is sample data being passed into the table. Pass in campersTest to view the sample data and emptyTest to view the empty state
3. To test the 3 dot button, it currently only console logs "3 dot button pressed!"

* Does the component adequately match the designs?
* Code quality and placement

## Checklist


frontend/src/components/pages/camps/CamperDetailsBadges.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+import { Camper } from "../../../types/CamperTypes";
+
+export const CamperDetailsBadge = ({
+    customIcon,
+    icon,
+    description,
+    color,
+  }: {
+    customIcon: boolean;
```
dhmi19, line 26: This works too, but another way to do this would be to make this field `customIconPath` an optional string. That way, when the string is present you use the local image and if there is no string, you can use the icon passed in into icon. The benefit of this would be that you have defined types and aren't using any. Up to you though, just a stylistic preference imo

frontend/src/components/pages/camps/CamperDetailsBadges.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <Text>{description}</Text>
+      </HStack>
+    );
+  };
+  
+  export const CamperDetailsBadgeGroup = ({ camper }: { camper: Camper }) => {
+    return (
+      <Container width="-webkit-fit-content" marginStart="0px">
+        <HStack>
+          {camper.latePickup && camper.earlyDropoff && (
```
dhmi19, line 53: I think there are 3 states for EDLP:
1. Late pickup AND NOT Early Dropoff
2. Early Dropoff AND NOT Late pickup
3. BOTH Late pickup and early dropoff 

image

dhmi19: I think we might just remove the third status though so just hold off on this change for now

frontend/src/components/pages/camps/CampersTable.tsx
```diff
@@ -0,0 +1,190 @@
...
+                <SearchIcon color="gray.300" />
+              </InputLeftElement>
+              <Input
+                size="md"
+                enterKeyHint="enter"
+                onChange={(e) => setSearch(e.target.value)}
+                placeholder="Search by camper name..."
+              />
+            </InputGroup>
+            <CampersTableOverviewIconGroup
```
dhmi19, line 110: The icon buttons above the camper rows are meant to be filters and the counts in each of these filters needs to be dynamically changing based on the search results. I think you would need to pass in `tableData` to the icons so that you can keep updating the numbers in the icons as the user searches in the search bar. You will also need a way to filter results when a user clicks on the icon buttons. As a result, the counts of the other filters will change accordingly. 

image
All the filters besides the ( person 4/20) filter should be updated with the new counts upon search

frontend/src/components/pages/camps/CampersTableNavBar.tsx
```diff
@@ -0,0 +1,138 @@
...
+        colorScheme="green"
+        outline="0"
+      >
+        <TabList>
+          <Tab whiteSpace="nowrap">Registration List</Tab>
+          <Tab>Waitlist</Tab>
+        </TabList>
+        <TabPanels width="100%" marginTop="16px">
+          <TabPanel padding="0">
+            <CampersTable campers={campersTest} campCapacity={campCapacity} />
```
dhmi19, line 128: You would pass in the selected session's camper list and camp capacity here right? The table would need to render the data for the selected session. Could you leave a TODO comment to make this dependant on the session data here?

frontend/src/components/pages/camps/CampersTableOverviewIconGroup.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+import { Camper } from "../../../types/CamperTypes";
+
+const TableOverviewIcon = ({
+  customIcon,
+  icon,
+  description,
+  color,
+}: {
+  customIcon: boolean;
```
dhmi19, line 22: my comment about using an optional string prop here would apply here too I think

frontend/src/components/pages/camps/CampersTableOverviewIconGroup.tsx
```diff
@@ -0,0 +1,104 @@
...
+  icon: any;
+  description: string;
+  color: string;
+}) => {
+  return (
+    <HStack
```
dhmi19, line 28: These are filters, so they will need an onclick function which would further filter out the results of the table. See the following specs on the dev handoff page:
image

Note that we don't need the camera and laptop filters (as well as the statuses for them in the table rows)

dhmi19 (changes requested):
Coming along well Ryan! I think we wanted the icons on the table (next to the search bar) to be toggle buttons and act like filters. This should be part of this ticket. Besides that, the rest looks good overall

annechung (approved):
tested and looks good

PR #100, opened 2022-08-31
feat: Implement access control table change status function

AmandaMisjuwar:
## Notion ticket link

[Access Management: Change user status](https://www.notion.so/uwblueprintexecs/Access-Management-Change-user-status-549c554ea8f14db5a7dee9ab43de28aa)

## Implementation description
* Added onClick handler to the 3 dots symbol for each row to show a confirmation modal popup and allows user to mark a user in the table as inactive or active

## Steps to test
1. Log in as an admin and go to the Access Control table
2. Choose a user that is active, and mark them as inactive. They should properly be updated in MongoDB and a success toast should be displayed.
3. Choose a user that is inactive, and mark them as active. This should also be updated in MongoDB and a success toast should be displayed.
4. Make sure the designs accurately match

* Coding practices
* Make sure components match designs

## Checklist

### DIFF
--- a/frontend/src/components/pages/AccessControl/index.tsx
+++ b/frontend/src/components/pages/AccessControl/index.tsx
@@ -1,6 +1,7 @@
 import { SearchIcon } from "@chakra-ui/icons";
 import { FaEllipsisV } from "react-icons/fa";
 import {
+  Button,
   Container,
   HStack,
   IconButton,
@@ -21,6 +22,7 @@ import {
   Th,
   Thead,
   Tr,
+  useDisclosure,
   useToast,
 } from "@chakra-ui/react";
 import React from "react";
@@ -28,6 +30,7 @@ import UserStatusLabel from "./UserStatusLabel";
 import UserAPIClient from "../../../APIClients/UserAPIClient";
 import { UserResponse } from "../../../types/UserTypes";
 import { Role } from "../../../types/AuthTypes";
+import ConfirmStatusChangeModal from "./ConfirmStatusChangeModal";
 
 const AccessControlPage = (): JSX.Element => {
   enum Filter {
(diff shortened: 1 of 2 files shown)


frontend/src/components/pages/AccessControl/index.tsx
```diff
@@ -202,11 +282,24 @@ const AccessControlPage = (): JSX.Element => {
...
+                      variant=""
+                    />
+                  </PopoverTrigger>
+                  <PopoverContent width="inherit">
+                    <PopoverBody
+                      as={Button}
+                      bg="background.white.100"
+                      onClick={(e) => handleStatusChangePopoverTrigger(user)}
+                    >
+                      Mark as {user.active ? Filter.INACTIVE : Filter.ACTIVE}
```
annechung, line 299: nit: can we define separate consts/enums for lowercase active/inactive strings shown here and in the modal? so we can show "Mark as active" instead of "Mark as Active"

annechung, line 299: nit: I think the end of the popover should be aligned with the ellipsis icon, and the text should be styled a bit differently?
Current:
image

Design:
image

annechung (approved):
tested and looks good! just a few comments about styling

frontend/src/components/pages/AccessControl/index.tsx
AmandaMisjuwar: Made it to match the design styling but the font still looks rly big because we needa scale down the font sizes as well

AmandaMisjuwar: Gonna merge anyway and we'll just need to update the textStyles when we get the updated sizes

annechung: yup yup sounds good

PR #108, opened 2022-09-22
fix: get users

annechung:
## Implementation description
- Currently when calling the get users endpoint, the entire request fails when one user is invalid (i.e. lookup in Firebase auth fails). Instead of failing the entire request we want to return the valid users.
- Adding some nullchecks where the UI was crashing
- Adding linting checks on dev branch

## Steps to test
1. Remove the `authId` field on a dummy user in the db to make this user invalid (or make a new entry leaving out the `authId` field).
2. Make a GET request to /users (navigating to a camp details page when you're signed in as an admin is an easy way to do this, can use http://localhost:3000/admin/camp/63139c7bc3d7b55b44a01531). 
3. Check in the network tab that the request to /users was successful and has returned users.

* 

## Checklist

### DIFF
--- a/frontend/src/components/pages/CampOverview/index.tsx
+++ b/frontend/src/components/pages/CampOverview/index.tsx
@@ -38,7 +38,7 @@ const CampOverviewPage = (): JSX.Element => {
     volunteers: "",
     campPhotoUrl: "",
   });
-  const numSessions = camp.campSessions.length;
+  const numSessions = camp.campSessions?.length;
 
   useEffect(() => {
     const getCamp = async () => {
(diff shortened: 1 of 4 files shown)


frontend/src/components/pages/CampOverview/index.tsx
```diff
@@ -38,7 +38,7 @@ const CampOverviewPage = (): JSX.Element => {
     volunteers: "",
     campPhotoUrl: "",
   });
-  const numSessions = camp.campSessions.length;
+  const numSessions = camp.campSessions?.length;
```
dhmi19, line 41: Is this a fix related to this bug ticket or just a fix in general?

annechung: fix in general!

dhmi19 (approved):
lgtm

PR #110, opened 2022-09-27
Camp Overview: Edit Camper Information

dhmi19:
## Notion ticket link

[Camp Overview: Edit Camper Information](https://www.notion.so/uwblueprintexecs/Task-Board-F22-9e03bf9f7c4343f09a8a5c16dc48a12a?p=e1dd8cd472ef4ad6a2de3cf1026e01fa&pm=c)

## Implementation description
* Updated the updateCampersByID backend route + function to only update the fields of a camper that can be updated by the client
* Made a folder called `EditCamperModal` which has all the files needed for the edit camper modal functionality. 
* Made a type for editCamperModal which is used by the different files to pass around props 

## Steps to test
1. Go to a camp overview page (http://localhost:3000/admin/camp/63139c7bc3d7b55b44a01531) and click on the hamburger icon next to a registered camper (use mine if you want)
2. Click on the edit button and a modal with all the current info should appear 
3. Try to change all the fields to make sure the results get updated as expected
4. The page should reload after 3 seconds (this might not be the exact functionality we want but the alternative was taking too long to implement for now)

* I removed a lot of code from the backend so just make sure that looks ok 
* Currently, I am refreshing the page after 3 seconds so that the new camper info is loaded after edit. Not sure if this is the best functionality. Ideally we use a global state but I didn't have time to implement that

## Checklist


frontend/src/components/pages/CampOverview/CampDetails.tsx
```diff
@@ -88,8 +88,8 @@ const CampDetails = ({ camp, setCamp }: CampDetailsProps): JSX.Element => {
     setCamp({ ...camp, campCounsellors });
   };
 
-  const handleVolunteerChange = (e: any) => {
-    setCamp({ ...camp, volunteers: e.target.value });
+  const handleVolunteerChange = (newVolunteers: string) => {
```
dhmi19, line 91: Not sure why this change got pulled in- it's not part of my changes. Should I remove this from the PR?

frontend/src/components/pages/CampOverview/CampDetails.tsx
```diff
@@ -88,8 +88,8 @@ const CampDetails = ({ camp, setCamp }: CampDetailsProps): JSX.Element => {
...
+    setCamp({ ...camp, volunteers: newVolunteers });
```
dhmi19, line 92: likewise for all the changes in this file

frontend/src/components/pages/CampOverview/CampDetails.tsx
twilkhoo: Hmm its not in dev, is this one of your own changes from another ticket in progress? If so, I guess it'll be more organized to remove it.

frontend/src/types/CamperTypes.ts
```diff
@@ -1,3 +1,11 @@
+export type EmergencyContact = {
```
twilkhoo, line 1: nit: is there a reason why this is "EmergencyContact" and not just "Contact"?

backend/typescript/middlewares/validators/camperValidators.ts
```diff
@@ -211,34 +211,6 @@ export const updateCamperDtoValidator = async (
   if (req.body.allergies && !validatePrimitive(req.body.allergies, "string")) {
     return res.status(400).send(getApiValidationError("allergies", "string"));
   }
-  if (!Array.isArray(req.body.earlyDropoff)) {
```
twilkhoo, line 214: I'm a bit confused can you explain why we're deleting all of this? Is it just not necessary?

frontend/src/components/pages/CampOverview/EditCamperModal/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+    if (isEditCamperSuccess) {
+      toast({
+        description: `${camper.firstName} ${camper.lastName}'s information has been updated`,
+        status: "success",
+        variant: "subtle",
+        duration: 3000,
+      });
+
+      setTimeout(() => {
+        window.location.reload();
```
twilkhoo, line 95: I know this is temporary, but if you want a non-reload solution you can check out what I did https://github.com/uwblueprint/focus-on-nature/pull/129
basically I'm triggering a refetch by changing some state in the useEffect's dependency array and I'm sure you could do that here too, it's still not ideal but it's better than an entire page reload (even temporarily).

frontend/src/components/pages/CampOverview/EditCamperModal/EditCamperFormResponseSection.tsx
```diff
@@ -0,0 +1,170 @@
...
+    question: string,
+    answer: string | undefined,
+    options: string[],
+    questionId: string,
+    isRequired: boolean,
+  ) => {
+    return (
+      <FormControl key={questionId} isRequired={isRequired}>
+        <FormLabel>
+          <b>{question}</b>
```
twilkhoo, line 41: Whereever you use `<b>`, I would reccomend using Chakra's `<Text>` element instead, and passing in some props regarding weight just so we stick to Chakra whereever possible.

frontend/src/components/pages/CampOverview/EditCamperModal/EditCamperFormResponseSection.tsx
```diff
@@ -0,0 +1,170 @@
...
+                onChange={() => {
+                  updateFormResponse(question, option);
+                }}
+              >
+                {option}
+              </Radio>
+            ))}
+          </Stack>
+        </RadioGroup>
+        <br />
```
twilkhoo, line 58: Instead of `<br>` add lower padding to the element above (in this case `<RadioGroup>`, same comment whereever else you use `<br>`s to avoid pure HTML elements.

frontend/src/APIClients/CamperAPIClient.ts
```diff
@@ -2,9 +2,38 @@ import { BEARER_TOKEN } from "../constants/AuthConstants";
...
+  camperData: EditCamperInfoFields,
+): Promise<Array<Camper>> => {
+  try {
+    const body = { camperIds: id, ...camperData };
+
+    // Can't send Map<string, string> over HTTP so need to convert to Object before
+    if (camperData.formResponses) {
+      body.formResponses = Object.assign(
+        {},
+        ...Array.from(camperData.formResponses.entries()).map(([k, v]) => ({
```
twilkhoo, line 21: nit: Instead of k, v you could use something more descriptive like question, response?

twilkhoo (commented):
Most of this looks amazing I left some comments regarding some HTML decisions and the window reloading. One thing I should address though is that testing your branch locally, the modal shows up perfectly and I'm able to edit whatever I need to, but after saving and refreshing the page, the changes don't persist. I'm guessing this works on your end (hence the PR) but just wanted to confirm/doublecheck.

frontend/src/types/CamperTypes.ts
dhmi19: we refer to contacts as Emergency Contacts throughout our product- or at least that's the terminology we used to begin with so I wanted to keep it consistent

backend/typescript/middlewares/validators/camperValidators.ts
dhmi19: these fields can't be edited once the camper is registered. I took them out of the `UpdateCamperDTO` object too. Currently, the validator was looking for those fields but since they can't be edited, it doesn't make sense to include them.

frontend/src/components/pages/CampOverview/EditCamperModal/index.tsx
dhmi19: That's completely fair but to refresh the state, I would need to change some dependancy in the CampOverview's useEffect so that it can get all the camp data again. I think that could be a little messy and since we're gonna replace with global state anyways, I think it's fine as is for now just to get the ticket moving. 

Good call out though- hadn't thought of that for this use case

frontend/src/components/pages/CampOverview/CampDetails.tsx
dhmi19: reverted it back to dev's file

dhmi19:
> 

Regarding the update not working: 

Our data is inconsistent in the database again (really need to fix it now since it's a glaring issue affecting everyone). This time, the questions that were under the camper's info did not match the questions that the camp model had (this should never actually happen when we stop playing with the db so often). Hence, the update kept failing. I tested it again with the correct questions setup for the camper and it worked:

https://user-images.githubusercontent.com/57159970/196326019-0b18ec01-392f-4010-a0d7-5fc669f6e9eb.mov

Note: updating the MCQ didn't work on the demo but it's probably because I didn't set up the db correctly. I have tested it before we changed the data for the guelph camp

frontend/src/components/pages/CampOverview/EditCamperModal/EditCamperFormResponseSection.tsx
dhmi19: done

frontend/src/APIClients/CamperAPIClient.ts
dhmi19: done

frontend/src/components/pages/CampOverview/EditCamperModal/EditCamperFormResponseSection.tsx
dhmi19: done

frontend/src/components/pages/CampOverview/EditCamperModal/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+        <ModalBody overflowY="scroll">
+          <EditCamperModalForm
+            formQuestions={formQuestions}
+            formStateVariables={stateVariables}
+            setStateFuncs={setStateFuncs}
+          />
+        </ModalBody>
+
+        <ModalFooter>
+          <Button onClick={editCamperOnClose} variant="editModalCancel">
```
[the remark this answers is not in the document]
dhmi19: done

e-wai:
@dhmi19 do you mind rebasing your changes? I think there are changes that aren't yours

backend/typescript/middlewares/validators/camperValidators.ts
twilkhoo: Ah now that makes sense to me.

frontend/src/components/pages/CampOverview/EditCamperModal/index.tsx
twilkhoo: Sounds good.

twilkhoo:
> > 
> 
> Regarding the update not working:
> 
> Our data is inconsistent in the database again (really need to fix it now since it's a glaring issue affecting everyone). This time, the questions that were under the camper's info did not match the questions that the camp model had (this should never actually happen when we stop playing with the db so often). Hence, the update kept failing. I tested it again with the correct questions setup for the camper and it worked:
> 
>  Edit.Camper.Info.WIP.mov 
> Note: updating the MCQ didn't work on the demo but it's probably because I didn't set up the db correctly. I have tested it before we changed the data for the guelph camp

Thanks for the vid, everything else looks good now! I'm giving the LGTM on my end, but be sure to confirm everything with Emma too regarding her comments (rebasing, etc.)

twilkhoo (approved):
I'm giving the LGTM on my end, but be sure to confirm everything with Emma too regarding her comments (rebasing, etc.), nice job on this!

frontend/src/APIClients/CamperAPIClient.ts
```diff
@@ -2,9 +2,40 @@ import { BEARER_TOKEN } from "../constants/AuthConstants";
...
+
+    // Can't send Map<string, string> over HTTP so need to convert to Object before
+    if (camperData.formResponses) {
+      body.formResponses = Object.assign(
+        {},
+        ...Array.from(camperData.formResponses.entries()).map(
+          ([question, response]) => ({
+            [question]: response,
+          }),
+        ),
```
e-wai, line 25: nit: is this different from `new JSONObject(camperData.formResponses);`?

frontend/src/components/pages/CampOverview/EditCamperModal/EditCamperFormResponseSection.tsx
```diff
@@ -0,0 +1,175 @@
...
+              : null;
+          case "Text":
+            return renderText(
+              question,
+              camperAnswer,
+              formQuestion.id,
+              formQuestion.required,
+            );
+          default:
+            return null;
```
e-wai, line 168: If this is exhaustive, we should probably do some sort of error here?

frontend/src/components/pages/CampOverview/EditCamperModal/EditCamperFormResponseSection.tsx
```diff
@@ -0,0 +1,175 @@
...
+        );
+        const camperAnswer =
+          camperAnswerIndex !== -1
+            ? Object.values(camperResponses)[camperAnswerIndex]
+            : "";
+
+        switch (type) {
+          case "MultipleChoice":
+            return formQuestion.options
+              ? renderMCQ(
```
e-wai, line 142: nit: is ts smart enough that it understands if you put a null check before the switch case (in which it returns null_, and so then you don't have to null check each case?

frontend/src/components/pages/CampOverview/EditCamperModal/EditCamperFormResponseSection.tsx
```diff
@@ -0,0 +1,175 @@
...
+    <VStack align="start">
+      {formQuestions.map((formQuestion) => {
+        const { question, type } = formQuestion;
+        const camperAnswerIndex = Object.keys(camperResponses).indexOf(
+          question,
+        );
+        const camperAnswer =
+          camperAnswerIndex !== -1
+            ? Object.values(camperResponses)[camperAnswerIndex]
+            : "";
```
e-wai, line 137: nit: extract to own func

frontend/src/components/pages/CampOverview/EditCamperModal/EditCamperModalForm.tsx
```diff
@@ -0,0 +1,125 @@
...
+              </Text>
+            </FormLabel>
+            <Input
+              onChange={(event) => {
+                setStateFuncs.setAge(parseInt(event.target.value, 10));
+              }}
+              value={formStateVariables.age}
+            />
+          </FormControl>
+        </VStack>
```
e-wai, line 76: These are difficult to read, can we create a component for the `VStack` and children, and then pass in appropriate parameters for each response?

frontend/src/components/pages/CampOverview/EditCamperModal/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+      toast({
+        description: `${camper.firstName} ${camper.lastName}'s information has been updated`,
+        status: "success",
+        variant: "subtle",
+        duration: 3000,
+      });
+
+      setTimeout(() => {
+        window.location.reload();
+      }, 3000);
```
e-wai, line 97: Can we move down from 3 secs? Even as the brute force approach - maybe we can't guarantee that we'll cover everyone's network speed, but I suspect this is going to feel like an error for most reasonable users

frontend/package.json
```diff
@@ -24,6 +24,7 @@
     "@types/react": "^17.0.0",
     "@types/react-dom": "^17.0.0",
     "@types/react-jsonschema-form": "^1.7.4",
+    "@types/react-select": "^5.0.1",
```
e-wai, line 27: Hmm it's late so I may be missing something, where is this used?

frontend/src/components/pages/CampOverview/EditCamperModal/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+    <Modal
+      isOpen={editCamperModalIsOpen}
+      onClose={editCamperOnClose}
+      preserveScrollBarGap
+    >
+      <ModalOverlay />
+      <ModalContent
+        w="1000px"
+        minW="1000px"
+        maxW="1000px"
```
e-wai, line 118: Do we need the initial width attribute?

frontend/src/theme/components/Button.ts
```diff
@@ -16,6 +16,21 @@ const Button = {
...
+    },
+    editModalSave: {
+      background: "primary.green.100",
+      padding: "12px, 25px, 12px, 25px",
+      radius: "6px",
+      width: "91px",
+      height: "48px",
+      color: "text.white.100",
+    },
   },
```
e-wai, line 34: nit: not a huge fan of making these theme components, since we're not reusing them and most of this isn't generally applicable styling?

frontend/src/components/pages/CampOverview/EditCamperModal/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+            marginRight="12px"
+          >
+            Cancel
+          </Button>
+          <Button
+            variant="editModalSave"
+            onClick={() => {
+              onSaveClicked();
+            }}
+          >
```
e-wai, line 147: nit: `onClick={onSaveClicked}`

e-wai (changes requested):
Awesome, awesome changes 🤩 Couple of comments, don't think it's anything structural. The only thing I'll block on is the `react-select` component

Also can you add a screenshot so we expect it to look like?

frontend/src/APIClients/CamperAPIClient.ts
dhmi19: I'm not sure which function you are referring to. TS didn't recognize it and I can't seem to find it online. Could you link it for me?

frontend/src/components/pages/CampOverview/EditCamperModal/EditCamperFormResponseSection.tsx
dhmi19: Yeah so this should never really be the case since the value is retrieved from the backend. What kinda error handling were you thinking?

frontend/src/components/pages/CampOverview/EditCamperModal/EditCamperFormResponseSection.tsx
dhmi19: The null check is not in each case, so we would have to break it up for the ones that need options and the TEXT type which doesn't. I can replace the ternary with `&&` to make it a bit cleaner though

frontend/src/components/pages/CampOverview/EditCamperModal/EditCamperFormResponseSection.tsx
dhmi19: done

frontend/src/components/pages/CampOverview/EditCamperModal/EditCamperModalForm.tsx
dhmi19: done

frontend/package.json
dhmi19: I think I got an indication from VS code to update versions at some point. I reverted the file to dev and everything worked, so I made changed it back to dev's version

frontend/src/components/pages/CampOverview/EditCamperModal/index.tsx
dhmi19: removed

frontend/src/theme/components/Button.ts
dhmi19: shfited to inline styling

frontend/src/components/pages/CampOverview/EditCamperModal/index.tsx
dhmi19: done

frontend/src/components/pages/CampOverview/EditCamperModal/index.tsx
dhmi19: changed it to 2 seconds

frontend/src/components/pages/CampOverview/EditCamperModal/EditCamperFormResponseSection.tsx
e-wai: Hmmm I'm not sure what the patterns are here: logging, alert, throw error (don't remember if it's swallowed once deployed?)?

frontend/src/components/pages/CampOverview/EditCamperModal/index.tsx
e-wai: How about something like an overall `sessionDidChange`? I suggest it in https://github.com/uwblueprint/focus-on-nature/pull/128/files# anyways

https://github.com/uwblueprint/focus-on-nature/blob/29c5d58a7f3726d57a5d823119ae6f488c566602/frontend/src/components/pages/CampOverview/index.tsx#L65-L74

e-wai (changes requested):
There's scroll behaviour when the modal is open, can we avoid that?

Screen Shot 2022-10-24 at 2 17 41 AM

frontend/src/APIClients/CamperAPIClient.ts
dhmi19: I tried that with the following code:
```
if (camperData.formResponses instanceof Map) {
      // body.formResponses = Object.assign(
      //   {},
      //   ...Array.from(camperData.formResponses.entries()).map(
      //     ([question, response]) => ({
      //       [question]: response,
      //     }),
      //   ),
      // );

      body.formResponses = Object.fromEntries(camperData.formResponses);
    }
```
But I get the following error:
```
(property) formResponses?: Map<string, string> | undefined
Type '{ [k: string]: string; }' is missing the following properties from type 'Map<string, string>': clear, delete, forEach, get, and 8 more.ts(2740)
```

frontend/src/components/pages/CampOverview/EditCamperModal/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+    hasPaid: camper.hasPaid,
+    firstName,
+    lastName,
+    age,
+    allergies,
+    specialNeeds,
+    formResponses,
+  };
+
+  const onSaveClicked = async () => {
```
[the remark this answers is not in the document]
dhmi19: addressed delay

e-wai (approved):
Yeah seems better to me, still think we can go down to 1000 ms or so for refetch delay unless you've had network problems with that?

dhmi19:
> Yeah seems better to me, still think we can go down to 1000 ms or so for refetch delay unless you've had network problems with that?

Which delay are you referring to? I think the refetch call is made immediately after we get a success now no?

e-wai:
@dhmi19 Oh my bad, read wrong -- looks good

PR #111, opened 2022-09-27
Create DELETE endpoint for multiple camp sessions

e-wai:
## Notion ticket link
[Camp Overview: Delete Session + Change Session Capacity](https://www.notion.so/uwblueprintexecs/8902e20b0cae4db2b4776259238f8a4e?v=4cf916ddabfc4438a2685c0990a4fc57&p=6d18cb517d61447ab691f8a7d37978f6&pm=c)

## Implementation description
* Note that implementation is different that `deleteCampersById`

## Steps to test
1. Make a DELETE call with existing camp sessions
3. Make a DELETE call with non-existing camp sessions

* Nothing specific, is the approach readable and does it makes sense?
  * better approach would use transactions

## Checklist

### DIFF
--- a/backend/typescript/services/implementations/campService.ts
+++ b/backend/typescript/services/implementations/campService.ts
@@ -12,8 +12,6 @@ import {
   UpdateCampDTO,
   CreateCampSessionsDTO,
   FormQuestionDTO,
-  CamperDTO,
-  WaitlistedCamperDTO,
 } from "../../types";
 
 import ICampService from "../interfaces/campService";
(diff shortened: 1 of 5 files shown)


backend/typescript/services/implementations/campService.ts
```diff
@@ -852,6 +850,111 @@ class CampService implements ICampService {
...
+          throw new Error(
+            `Not all camp sessions with IDs in [${campSessionIds}] were found in camp with ID ${campId}.`,
+          );
+        }
+      });
+
+      const oldCampSessions = camp.campSessions; // clone the full array of campers for rollback
+
+      try {
+        // delete camper IDs from the array of campers in the camp session
```
jjjding2, line 897: Only campSessions seem to be touched in this block of code

backend/typescript/services/implementations/campService.ts
```diff
@@ -852,6 +850,111 @@ class CampService implements ICampService {
...
+            "MongoDB campSessionIdsthat could not be restored in the camp Session=",
+            campSessionIds,
+          ];
+          Logger.error(errorMessage.join(" "));
+        }
+      }
+
+      // deleting campers from the camper table
+      try {
+        await MgCampSession.deleteMany({
```
jjjding2, line 927: I think we need to delete the campers by their ids if they were contained in one of the delete campSessions

jjjding2 (commented):
Some confusion about how campers are handled in this delete endpoint – it doesn't look like campers are ever explicitly deleted (only campSessions) so the inline comments were confusing.

camper ids for a given campSession can be found in the `campers` attribute

backend/typescript/services/implementations/campService.ts
e-wai: Yeah it's a good callout. Will revisit - some inclination to use another service to do this work, but I think that would violate the idea of single transaction?

backend/typescript/services/implementations/campService.ts
```diff
@@ -852,6 +850,128 @@ class CampService implements ICampService {
...
+              $in: campSessionIds,
+            },
+          },
+        });
+      } catch (mongoDbError: unknown) {
+        const errorMessage = [
+          `Failed to delete all campers belonging to camp sessions ${campSessionIds} Reason =`,
+          getErrorMessage(mongoDbError),
+        ];
+        Logger.error(errorMessage.join(" "));
```
e-wai, line 963: Do we want to recreate the campers later?

backend/typescript/services/implementations/campService.ts
```diff
@@ -852,6 +850,128 @@ class CampService implements ICampService {
...
+          `Could not find all camp sessions with IDs in [${campSessionIds}] and belonging to camp with ID ${campId}.`,
+        );
+      }
+
+      const camp: Camp | null = await MgCamp.findById(campId);
+      if (!camp) {
+        throw new Error(`Camp with campId ${campId} not found.`);
+      }
+
+      // This is an edge case, where the camp session has a campId but the the session does not appear on that camp
```
dhmi19, line 881: why would this case ever be there?

backend/typescript/services/implementations/campService.ts
dhmi19: @e-wai you implemented this change below right or did I miss something?

dhmi19 (approved):
Few minor comments but looks good overall

backend/typescript/services/implementations/campService.ts
```diff
@@ -852,6 +850,128 @@ class CampService implements ICampService {
...
+      // deleting camp sessions from the camp session table
+      try {
+        await MgCampSession.deleteMany({
+          _id: {
+            $in: campSessionIds,
+          },
+        });
+      } catch (mongoDbError: unknown) {
+        try {
+          await MgCampSession.create(campSessionIds);
```
dhmi19, line 935: Does this repopulate the data that may have been lost when deletingthesession?

e-wai: It's been used previously for similar purposes, but wasn't sure how to explicitly reach this pathway. A bunch cleaner path would be to move to using transactions, thoughts?

PR #114, opened 2022-09-28
Create Manage Session Modal

e-wai:
## Notion ticket link

[Camp Overview: Delete Session + Change Session Capacity ](https://www.notion.so/uwblueprintexecs/Camp-Overview-Delete-Session-Change-Session-Capacity-6d18cb517d61447ab691f8a7d37978f6)

This PR handles creating the modal; followup pieces of work for this ticket:
1. Creating DELETE, PATCH endpoints
2. Attaching to endpoints with API clients

## Implementation description
* Create the modal component, passing `onOpen`, `onClose`, `isOpen` props to control open and close
* Create a table component within the modal
* Move some utility methods to `CampUtils`

## Steps to test
1. Go to a camp page, such as http://localhost:3000/admin/camp/630a63450eb59852f9d01b1e, click the new "Manage Sessions" modal and play around. There's no actual database interactions... look at delete button (and that everything is appropriately disabled), input fields
2. Check a camp page with no sessions, shouldn't see anything new

https://user-images.githubusercontent.com/40974372/192941408-3037a95e-d2ba-4593-8369-44842e9d4267.mov

* Does it make sense to use `Map` and `Set` respectively? Array is slow to search so this makes insertion/deletion/contains checks easier, and all of the above require O(n) time for changes. Admittedly, the scale here is like 4-5, so is this significantly harder to read?
* Naming questions
* There's a ton of business logic just hanging around, does it live in the appropriate place?
* Any styling notes (may be put into separate PR)

## Checklist

### DIFF
--- a/frontend/src/components/pages/CampOverview/ManageSessions/ManageSessionsModal.tsx
+++ b/frontend/src/components/pages/CampOverview/ManageSessions/ManageSessionsModal.tsx
@@ -19,7 +19,7 @@ interface ManageSessionsModalProps {
   sessions: Array<ManageCampSessionDetails>;
   onSaveChanges: (
     deletedSessions: Set<string>,
-    updatedCapacities: Map<string, number | undefined>,
+    updatedCapacities: Map<string, string>,
   ) => void;
   isOpen: boolean;
   onClose: () => void;
@@ -37,7 +37,7 @@ const ManageSessionsModal = ({
     Set<string>
   >(new Set());
   const [updatedSessionIdCapacity, setUpdatedSessionIdCapacity] = useState<
-    Map<string, number | undefined>
+    Map<string, string>
   >(new Map());
 
   const closeModalAndClearChanges = () => {
@@ -49,7 +49,7 @@ const ManageSessionsModal = ({
   return (
     <Modal isOpen={isOpen} onClose={closeModalAndClearChanges} isCentered>
       <ModalOverlay />
-      <ModalContent pt={5} minW="800px">
+      <ModalContent pt={5} minW="1000px" maxH="600px">
         <ModalHeader ml={6} textStyle="heading">
           <Text>Manage Camp Sessions</Text>
         </ModalHeader>
@@ -69,9 +69,7 @@ const ManageSessionsModal = ({
                   new Map(
                     updatedSessionIdCapacity.set(
                       updatedSession.id,
-                      updatedCapacity
-                        ? parseInt(updatedCapacity, 10)
-                        : undefined,
+                      updatedCapacity,
                     ),
                   ),
                 );
@@ -99,7 +97,6 @@ const ManageSessionsModal = ({
             >
               <Text>Cancel</Text>
             </Button>
-            {/* TODO implement onSave */}
             <Button
               colorScheme="green"
               onClick={() => {
--- a/frontend/src/components/pages/CampOverview/ManageSessions/ManageSessionsModal.tsx
+++ b/frontend/src/components/pages/CampOverview/ManageSessions/ManageSessionsModal.tsx
@@ -1,0 +1,119 @@
+import React, { useState } from "react";
+import {
+  Button,
+  HStack,
+  Modal,
+  ModalBody,
+  ModalContent,
+  ModalFooter,
+  ModalHeader,
+  ModalOverlay,
+  Text,
+} from "@chakra-ui/react";
+import { ManageCampSessionDetails } from "../../../../types/CampsTypes";
+import ManageSessionsModalTable from "./ManageSessionsModalTable";
+
+interface ManageSessionsModalProps {
+  campStartTime: string;
+  campEndTime: string;
+  sessions: Array<ManageCampSessionDetails>;
+  onSaveChanges: (
+    deletedSessions: Set<string>,
+    updatedCapacities: Map<string, number | undefined>,
+  ) => void;
+  isOpen: boolean;
+  onClose: () => void;
+}
+
+const ManageSessionsModal = ({
+  campStartTime,
+  campEndTime,
+  sessions,
+  onSaveChanges,
+  isOpen,
+  onClose,
+}: ManageSessionsModalProps): JSX.Element => {
+  const [deletedCampSessionIds, setDeletedCampSessionIds] = useState<
+    Set<string>
+  >(new Set());
+  const [updatedSessionIdCapacity, setUpdatedSessionIdCapacity] = useState<
+    Map<string, number | undefined>
+  >(new Map());
+
+  const closeModalAndClearChanges = () => {
+    setDeletedCampSessionIds(new Set());
+    setUpdatedSessionIdCapacity(new Map());
+    onClose();
+  };
+
+  return (
+    <Modal isOpen={isOpen} onClose={closeModalAndClearChanges} isCentered>
+      <ModalOverlay />
+      <ModalContent pt={5} minW="800px">
+        <ModalHeader ml={6} textStyle="heading">
+          <Text>Manage Camp Sessions</Text>
+        </ModalHeader>
+        <ModalBody>
+          {sessions.length === 0 ? (
+            <Text>No sessions to edit</Text>
+          ) : (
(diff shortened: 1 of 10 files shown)


e-wai:
**this colours are a little off, but that can be a fix PR

frontend/src/components/pages/CampOverview/ManageSessions/ManageSessionsModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const closeModalAndClearChanges = () => {
+    setDeletedCampSessionIds(new Set());
+    setUpdatedSessionIdCapacity(new Map());
+    onClose();
+  };
+
+  return (
+    <Modal isOpen={isOpen} onClose={closeModalAndClearChanges} isCentered>
+      <ModalOverlay />
+      <ModalContent pt={5} minW="800px">
```
dhmi19, line 52: I think the modal is supposed to be a fixed height of 1000px and width of 600px

frontend/src/components/pages/CampOverview/ManageSessions/ManageSessionsModalTable.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const sessionIsFull = (session: ManageCampSessionDetails): boolean =>
+    session.capacity === session.registeredCampers;
+
+  return (
+    <TableContainer>
+      <Table variant="simple">
+        <Thead>
+          <Tr>
+            <Th textTransform="none">
+              <Text textStyle="bodyRegular">Camp Session</Text>
```
dhmi19, line 79: Unless we decided to change these sizes, bodyRegular is supposed to be 18px I think. Is this something we need to change in textStyles file?
image

frontend/src/components/pages/CampOverview/ManageSessions/ManageSessionsModalTable.tsx
```diff
@@ -0,0 +1,157 @@
...
+                      <Tag colorScheme="orange" borderRadius="full" size="sm">
+                        Session Full
+                      </Tag>
+                    ) : (
+                      <Tag colorScheme="green" borderRadius="full" size="sm">
+                        Spots Available
+                      </Tag>
+                    )}
+                  </HStack>
+                  <Text textStyle="bodyRegular">
```
dhmi19, line 106: I think the text here is supposed to be a bit darker too. Might be the same change as the table headings above (they seem to have the same style on the figma)

frontend/src/components/pages/CampOverview/ManageSessions/ManageSessionsModalTable.tsx
```diff
@@ -0,0 +1,157 @@
...
+                >
+                  <HStack>
+                    <SessionRowTitle session={session} index={index} />
+                    {sessionIsFull(session) ? (
+                      <Tag colorScheme="orange" borderRadius="full" size="sm">
+                        Session Full
+                      </Tag>
+                    ) : (
+                      <Tag colorScheme="green" borderRadius="full" size="sm">
+                        Spots Available
```
dhmi19, line 102: Typo: Session available

frontend/src/components/pages/CampOverview/ManageSessions/ManageSessionsModalTable.tsx
```diff
@@ -0,0 +1,157 @@
...
+            return (
+              <Tr key={index}>
+                <Td
+                  layerStyle={deletedSessions.has(session.id) ? "disabled" : ""}
+                >
+                  <HStack>
+                    <SessionRowTitle session={session} index={index} />
+                    {sessionIsFull(session) ? (
+                      <Tag colorScheme="orange" borderRadius="full" size="sm">
+                        Session Full
```
dhmi19, line 98: Typo: Session full (no caps on Full)

frontend/src/utils/CampUtils.ts
```diff
@@ -34,3 +34,26 @@ export const campStatus = (camp: CampResponse): CampStatus => {
...
+  );
+
+  return `${startDate} - ${endDate}`;
+};
+
+export const getFormattedSessionDatetime = (
+  dates: Array<string>,
+  startTime: string,
+  endTime: string,
+): string => `${getFormattedDateString(dates)} | ${startTime} - ${endTime}`;
```
dhmi19, line 59: The modal you rendered doesn't show AM and PM. I can't see if this is an issue with data on the db or something you forgot in the code?

frontend/src/components/pages/CampOverview/ManageSessions/ManageSessionsModal.tsx
```diff
@@ -0,0 +1,119 @@
...
+  sessions,
+  onSaveChanges,
+  isOpen,
+  onClose,
+}: ManageSessionsModalProps): JSX.Element => {
+  const [deletedCampSessionIds, setDeletedCampSessionIds] = useState<
+    Set<string>
+  >(new Set());
+  const [updatedSessionIdCapacity, setUpdatedSessionIdCapacity] = useState<
+    Map<string, number | undefined>
```
dhmi19, line 40: Can we not avoid the undefined here? If the user removes the value in the field, can you remove session from this map or just populate with its existing value? Unless I missed something, I think it would be cleaner to avoid the undefined and keep this state purely the session and updated capacity

dhmi19 (changes requested):
I think it's pretty much fine. Just left some minor comments here and there. Addressing your concerns below:

1. Map and Set is fine by me. It's a little bit harder to read but I think I get what you're trying to do and it makes sense to use those DS. Left a comment about the undefined in the Map though- I think we can avoid it?

2. Naming seems fine to me

3. Given the current project structure, I think it's fine. If you really wanted to, you could extract some of the business logic in custom components (eg: make tags their own component where we can pass in state etc.), but I think it's fine

4. left some comments on styling (most of them are styling actually lol)

e-wai (commented):
fixed up some basic stuff, mostly the input value logic to revisit

frontend/src/utils/CampUtils.ts
e-wai: Quick guess would be backend? I don't see any parsing on the frontend :/

e-wai: Yeah backend

frontend/src/components/pages/CampOverview/ManageSessions/ManageSessionsModal.tsx
e-wai: ^ this was the change

dhmi19:
I don't think the edit session capacity is working? I tried to change session 1's capacity to 25 instead of 30 but it didn't seem to update. Also, you might want to force a reload on update so that the frontend gets updated. At least as a temporary fix before we introduce global state?

frontend/src/components/pages/CampOverview/ManageSessions/ManageSessionsModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+            <Text>No sessions to edit</Text>
+          ) : (
+            <ManageSessionsModalTable
+              campStartTime={campStartTime}
+              campEndTime={campEndTime}
+              sessions={sessions}
+              deletedSessions={deletedCampSessionIds}
+              updatedCapacities={updatedSessionIdCapacity}
+              onEditCapacity={(updatedSession, updatedCapacity) => {
+                // TODO check for duplicate calls (user types back in original capacity)
```
dhmi19, line 67: did you wanna do that in this pr?

dhmi19 (approved):
lgtm, just a TODO comment left over if you wanted to add that in this pr

frontend/src/components/pages/CampOverview/ManageSessions/ManageSessionsModal.tsx
e-wai: hmm I'll keep this out for now, there's 3 TODOs that I can think of:
- this one
- add tooltip for error state
- disable save button when invalid capacity

PR #115, opened 2022-09-30
Fix Edit Camper Entry Point

dhmi19:
added a selected camper state so everyone who needs the kebab menu ca…n use this to render their modals

## Notion ticket link

Not really a part of a ticket (maybe edit modal), but this is a quick push so that @twilkhoo and @uyxela can use the state to render their modals

## Implementation description
* added a state variable to set the selected camper on the campersTable file 
* Will use this state to pass info on all the modals related to that particular camper

## Steps to test
1. Go to http://localhost:3000/admin/camp/63139c7bc3d7b55b44a01531 
2. Try to edit a camper by going to the kebab menu -> edit 
3. You should see the edit modal with the selected camer's first name

* just want to make sure setting this state is the best way of doing this

## Checklist

### DIFF
--- a/frontend/src/components/pages/CampOverview/CampersTable/CampersTable.tsx
+++ b/frontend/src/components/pages/CampOverview/CampersTable/CampersTable.tsx
@@ -96,6 +96,10 @@ const CampersTable = ({
     specialNeeds: 0,
   });
 
+  const [selectedCamper, setSelectedCamper] = React.useState<
+    Camper | undefined
+  >();
+
   React.useMemo(() => {
     const tempDetailsCount = {
       earlyDropoff: 0,
@@ -219,7 +223,10 @@ const CampersTable = ({
                     maxWidth="32px"
                   >
                     <CampersTableKebabMenu
-                      editCamperFunc={editModalOnOpen}
+                      editCamperFunc={() => {
+                        setSelectedCamper(camper);
+                        editModalOnOpen();
+                      }}
                       viewDetailsFunc={() => {
                         console.log("Viewing Camper");
                       }}
@@ -238,11 +245,13 @@ const CampersTable = ({
 
           {/* Add the registered camper action modals here  */}
 
-          <EditCamperModal
-            camperId="hi"
-            editCamperModalIsOpen={editModalIsOpen}
-            editCamperOnClose={editModalOnClose}
-          />
+          {selectedCamper && (
+            <EditCamperModal
+              camper={selectedCamper}
+              editCamperModalIsOpen={editModalIsOpen}
+              editCamperOnClose={editModalOnClose}
+            />
+          )}
         </>
       ) : (
         <VStack pb="18px" pt="18px">
--- a/frontend/src/components/pages/CampOverview/EditCamperModal.tsx
+++ b/frontend/src/components/pages/CampOverview/EditCamperModal.tsx
@@ -10,15 +10,16 @@ import {
   ModalFooter,
   Button,
 } from "@chakra-ui/react";
+import { Camper } from "../../../types/CamperTypes";
 
 type EditCamperModalProps = {
-  camperId: string;
+  camper: Camper;
   editCamperModalIsOpen: boolean;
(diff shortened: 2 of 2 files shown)


frontend/src/components/pages/CampOverview/CampersTable/CampersTable.tsx
```diff
@@ -96,6 +96,8 @@ const CampersTable = ({
     specialNeeds: 0,
   });
 
+  const [selectedCamper, setSelectedCamper] = React.useState<Camper>();
```
e-wai, line 99: If we have a null state (you check that later), should we have the type `<Camper | null>`? or `<Camper | undefined>` maybe

frontend/src/components/pages/CampOverview/CampersTable/CampersTable.tsx
```diff
@@ -238,11 +243,13 @@ const CampersTable = ({
...
-            editCamperModalIsOpen={editModalIsOpen}
-            editCamperOnClose={editModalOnClose}
-          />
+          {selectedCamper ? (
+            <EditCamperModal
+              camper={selectedCamper}
+              editCamperModalIsOpen={editModalIsOpen}
+              editCamperOnClose={editModalOnClose}
+            />
+          ) : null}
```
e-wai, line 252: nit:
```
{
  selectedCamper &&
  <EditCamperModal ... />
}
```

frontend/src/components/pages/CampOverview/CampersTable/CampersTable.tsx
```diff
@@ -219,7 +221,10 @@ const CampersTable = ({
                     maxWidth="32px"
                   >
                     <CampersTableKebabMenu
-                      editCamperFunc={editModalOnOpen}
+                      editCamperFunc={() => {
```
e-wai, line 224: super super nit: Can we call these functions more along the lines of `onClick` -> eg. `onClickEditCamper`? and also abstract this to another method so that the TSX section is easier to parse

e-wai (changes requested):
quick mini changes?

frontend/src/components/pages/CampOverview/CampersTable/CampersTable.tsx
dhmi19: hmm not sure what you mean by "Abstract this to another method". You mean make a function which calls the editModalOpen function?

dhmi19: if so, i think it's fine and we can just merge it in? Not sure what the benefit would be tbh

e-wai: Yeah something like `onClickEditCamper={openModalForSelectedCamper}`, but it's mostly just the naming that I think is worth it. Your call, giving the ✅

e-wai (approved):

frontend/src/components/pages/CampOverview/EditCamperModal.tsx
```diff
@@ -28,7 +29,7 @@ const EditCamperModal = ({
       <ModalContent>
         <ModalHeader>Modal Title</ModalHeader>
         <ModalCloseButton />
-        <ModalBody>{camperId}</ModalBody>
+        <ModalBody>{camper.firstName}</ModalBody>
```
twilkhoo, line 32: Since camper can be undefined, would we need `camper?.firstName`? I know this is an incomplete modal but just wanted your thoughts since I'm doing the same thing in my modal.

twilkhoo (approved):
LGTM!

frontend/src/components/pages/CampOverview/EditCamperModal.tsx
dhmi19: I don't think the camper can be undefined. The props only allow for Camper type. I check the state of `selectedCamper` before passing it in as a prop

PR #123, opened 2022-10-05
feat: Filter camps by year

AmandaMisjuwar:
## Notion ticket link

[Filter camps by year](https://www.notion.so/uwblueprintexecs/GET-camps-filtering-by-year-f226bcab30cc4f04b38e848ca7f153c8)

## Implementation description
* Updated the get all camps endpoint by accepting a `year` parameter which will filter the returned camps by the specified year
* Slack discussion for year filtering logic: https://uwblueprint.slack.com/archives/C02R9QK34UB/p1664037904985479
* TLDR; every camp has a list of camp sessions. We return any camp that has camp sessions in the specified year. If the camp has no camp sessions AKA the list is empty, we use the camp's creation date instead.
* Changes include updating camps model to include auto-generated createdAt field, updating route validators to account for a year being passed in, and filtering logic

## Steps to test
1. Use Postman and make a query to the endpoint like so: `http://localhost:5000/camp?campYear=2022`
2. Play around with the year passed and make sure we have proper error checking i.e can't pass invalid year or letters
3. Make sure the camps being returned contain ALL the camp sessions in the specified year and that no incorrect camp or camp session is being returned in the response
4. [Lower priority] Test the case If a camp has no camp sessions assigned to it but it was created in the year specified.

#### Sample test cases
`year`: 2022
- Camp created in 2022, campSession in 2022 -> return
- Camp created in 2022, no campSession in 2022 -> return
- Camp created in 2023, no campSession in 2021 -> don't return
- Camp created in 2023, campSession in 2022 -> X (should not exist)

* Filtering logic and possible errors
* Making sure the data being returned is correct

## Checklist


backend/typescript/rest/campRoutes.ts
```diff
@@ -26,9 +27,10 @@ const fileStorageService: IFileStorageService = new FileStorageService(
 
 const campService: ICampService = new CampService(fileStorageService);
 /* Get all camps */
-campRouter.get("/", async (req, res) => {
+campRouter.get("/", getCampDtoValidator, async (req, res) => {
   try {
-    const camps = await campService.getCamps();
+    const { campYear } = req.query;
+    const camps = await campService.getCamps(parseInt(campYear as string, 10));
```
dhmi19, line 33: I tested this so I know it works but one of the concerns I had was if we don't pass in campYear then how does this parseInt ?

backend/typescript/rest/campRoutes.ts
```diff
@@ -26,9 +27,10 @@ const fileStorageService: IFileStorageService = new FileStorageService(
 
 const campService: ICampService = new CampService(fileStorageService);
 /* Get all camps */
-campRouter.get("/", async (req, res) => {
+campRouter.get("/", getCampDtoValidator, async (req, res) => {
   try {
-    const camps = await campService.getCamps();
+    const { campYear } = req.query;
```
dhmi19, line 32: If the URL doesn't include a year, should we default it to the current year? I can't imagine the frontend needing all the camps across all the years. The default view for the camp list should be the current year too

backend/typescript/models/camp.model.ts
```diff
@@ -35,134 +35,139 @@ export interface Camp extends Document {
...
       type: String,
-      required: false,
     },
-    city: {
+    dropoffPriceId: {
       type: String,
-      required: true,
     },
-    province: {
+    dropoffProductId: {
```
dhmi19, line 153: probably not related to this ticket but I saw that we have all of the following fields: 
1. dropoffFee
2. pickupFee
3. dropoffPriceId
4. dropoffProductId
5. pickupPriceId
6. pickupProductId

Are all these fields required for Stripe payments?

backend/typescript/models/camp.model.ts
```diff
@@ -35,134 +35,139 @@ export interface Camp extends Document {
...
+    },
+    pickupPriceId: {
+      type: String,
+    },
+    pickupProductId: {
       type: String,
-      required: true,
     },
-    postalCode: {
+    startTime: {
```
dhmi19, line 162: I'm a little confused about what you actually changed in the model. Could you just summarize the changes in the thread since the additions and removals don't seem to match up

backend/typescript/models/camp.model.ts
```diff
@@ -35,134 +35,139 @@ export interface Camp extends Document {
...
-  },
-  startTime: {
-    type: String,
-    required: true,
-  },
-  volunteers: {
-    type: String,
-  },
-});
+  { timestamps: true },
```
dhmi19, line 170: Does this correlate to the `createdAt` and `updatedAt` fields above? Also those fields are set to string not date. is that ok?

backend/typescript/services/implementations/campService.ts
```diff
@@ -68,6 +66,33 @@ class CampService implements ICampService {
         return [];
       }
 
+      if (year) {
+        // Filter out camps that have camp sessions in the desired year
+        /* eslint-disable no-param-reassign */
+        camps = camps.filter((camp) => {
+          // Go through every camp session in a camp. If it has the desired year, include it in the results.
```
dhmi19, line 73: can we reword this comment- it seems a bit misleading. Something like "go through every camp session in a camp and only include camps where ALL sessions are in the given year"

backend/typescript/services/implementations/campService.ts
```diff
@@ -68,6 +66,33 @@ class CampService implements ICampService {
...
+          /* eslint-disable no-param-reassign */
+          camp.campSessions = (camp.campSessions as CampSession[]).filter(
+            (campSession) => {
+              return campSession.dates.every((campDate) => {
+                const startCampYearTime = new Date(year, 0, 1).getTime();
```
dhmi19, line 78: I printed out what date this refers to and I got `Date: "2022-01-01T05:00:00.000Z" ` do we need to set the hour to 00?

backend/typescript/services/implementations/campService.ts
```diff
@@ -68,6 +66,33 @@ class CampService implements ICampService {
...
+                const endCampYearTime = new Date(year, 11, 31).getTime();
+                return (
+                  campDate.getTime() >= startCampYearTime &&
+                  campDate.getTime() <= endCampYearTime
```
dhmi19, line 82: Why not just check if the years are the same for the session date and given date? We could use the `getFullYear()` function?

backend/typescript/services/implementations/campService.ts
```diff
@@ -68,6 +66,33 @@ class CampService implements ICampService {
...
+                );
+              });
+            },
+          );
+
+          // After filtering all the campSessions that have the wanted year, return all camps that have campSessions in the year or were created in the year
+          return (
+            camp.campSessions.length > 0 ||
+            new Date(camp.createdAt).getFullYear() === year
```
dhmi19, line 91: ahh I see you used the function here. was there a reason you didn't use it for checking sessions?

dhmi19 (changes requested):
Tested the endpoint and it looks good. Didn't get around to doing a full on stress test but tried different invalid formats for the year, different year ranges etc and it worked well. Left a few comments mostly for clarifications

backend/typescript/rest/campRoutes.ts
e-wai: Hmm I think it makes sense to include all the camps when no year is specified, even if we don't currently have a year specified. For a GET "all" endpoint, I think it should behave that way be default unless an explicit parameter is included. For when we want the current year's camps, our frontend should have to explicitly say so

backend/typescript/services/implementations/campService.ts
```diff
@@ -68,6 +66,33 @@ class CampService implements ICampService {
...
 
+      if (year) {
+        // Filter out camps that have camp sessions in the desired year
+        /* eslint-disable no-param-reassign */
+        camps = camps.filter((camp) => {
+          // Go through every camp session in a camp. If it has the desired year, include it in the results.
+          /* eslint-disable no-param-reassign */
+          camp.campSessions = (camp.campSessions as CampSession[]).filter(
+            (campSession) => {
+              return campSession.dates.every((campDate) => {
```
e-wai, line 77: nit: Is it any easier to sort camp sessions by dates and check the first + last date? For the size that we're dealing with, it's not a huge performance thing but might be easier to read

backend/typescript/services/implementations/campService.ts
```diff
@@ -68,6 +66,33 @@ class CampService implements ICampService {
...
+                const endCampYearTime = new Date(year, 11, 31).getTime();
+                return (
+                  campDate.getTime() >= startCampYearTime &&
+                  campDate.getTime() <= endCampYearTime
+                );
+              });
+            },
+          );
+
+          // After filtering all the campSessions that have the wanted year, return all camps that have campSessions in the year or were created in the year
```
e-wai, line 88: Edge case: camp query for 2021, should we return a camp created in 2021 with no camp sessions in 2021 (eg. let's say it has 5 camp sessions in 2022)?

This comment + logic says yes, but from the Slack convo I thought we only wanted creation date if no camp sessions exist

e-wai (commented):
Just couple of questions from me, will let Dhruv give the final 👍

backend/typescript/models/camp.model.ts
```diff
@@ -35,134 +35,139 @@ export interface Camp extends Document {
   pickupPriceId: string;
   pickupProductId: string;
   volunteers: string;
+  createdAt: string;
+  updatedAt: string;
```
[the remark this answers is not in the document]
AmandaMisjuwar: I think we added it in case we'll need it in the future - it's autogenerated

backend/typescript/services/implementations/campService.ts
AmandaMisjuwar: The code for checking sessions was taken from the original PR that was open for this lol I can update it

backend/typescript/models/camp.model.ts
AmandaMisjuwar: I think @annechung would know for this one! Although I do think this is for Stripe payments

backend/typescript/models/camp.model.ts
AmandaMisjuwar: Sorry all I added was the `createdAt` and `updatedAt` fields, I think the rest of the fields might be from merges or linting fixes - I'll double check

backend/typescript/rest/campRoutes.ts
dhmi19: ok fair enough. I am good with it then

backend/typescript/models/camp.model.ts
dhmi19: oh ok, thanks

backend/typescript/services/implementations/campService.ts
dhmi19: yeah if it's the same thing, can you change it? I think this is cleaner. just verify that it works the same though

backend/typescript/models/camp.model.ts
AmandaMisjuwar: Yeah looks like it's all the same except createdAt, updatedAt, and the timestamps field

backend/typescript/models/camp.model.ts
AmandaMisjuwar: Updated it to Date based on https://codesource.io/learn-how-to-use-mongoose-timestamps/

backend/typescript/rest/campRoutes.ts
AmandaMisjuwar: I think it would just call the getCamps function with no value for the `year` argument. Since we will make the filter on the camps after calling all camps if a `year` exists, it just returns all the camps queried as required

backend/typescript/services/implementations/campService.ts
AmandaMisjuwar: For the sake of time will just move on with the current code since it works but will keep this in mind in case it does become an issue in the future!

backend/typescript/services/implementations/campService.ts
AmandaMisjuwar: Removed this line so shouldn't be an issue now!

backend/typescript/services/implementations/campService.ts
AmandaMisjuwar: done!

backend/typescript/services/implementations/campService.ts
AmandaMisjuwar: updated the logic to match Slack convo - will need to remove the check for undefined/null type once our test data all have the createdAt field

dhmi19 (approved):
lgtm!

PR #125, opened 2022-10-06
Feat/display registration link

jjjding2:
## Notion ticket link

[Camp Overview: Display registration link on top half + Copy interaction](https://www.notion.so/uwblueprintexecs/Task-Board-F22-9e03bf9f7c4343f09a8a5c16dc48a12a?p=9ee07dc64a114a048f3d9e3cec204dcc&pm=c)

## Implementation description
* Built a component that displays the registration link in a disabled input field
* Added a button that copies the link to clipboard, displaying a toast

image

## Steps to test
1. Go to a camp (i.e. http://localhost:3000/admin/camp/63139c7bc3d7b55b44a01531)
2. Copy the registration link
3. Test an inactive camp to see that the link is empty, button is disabled

* Behaviour for active/inactive camps, hardcoding for the link

## Checklist

### DIFF
--- a/frontend/src/components/pages/CampOverview/CampDetails.tsx
+++ b/frontend/src/components/pages/CampOverview/CampDetails.tsx
@@ -27,6 +27,7 @@ import {
   CampStatusColor,
 } from "../../../utils/CampUtils";
 import CampsAPIClient from "../../../APIClients/CampsAPIClient";
+import CampRegistrationLink from "./CampRegistrationLink";
 
 type CampDetailsProps = {
   camp: CampResponse;
@@ -110,6 +111,11 @@ const CampDetails = ({ camp, setCamp }: CampDetailsProps): JSX.Element => {
             </Tag>
           </HStack>
         </HStack>
+        <CampRegistrationLink
+          p={
+            camp.active ? `https://camps.focusonnature.ca/camp/${camp.id}` : ``
+          }
+        />
         <Grid
           templateColumns="repeat(5, 1fr)"
           gap={2}
--- a/frontend/src/components/pages/CampOverview/CampDetails.tsx
+++ b/frontend/src/components/pages/CampOverview/CampDetails.tsx
@@ -112,10 +112,10 @@ const CampDetails = ({ camp, setCamp }: CampDetailsProps): JSX.Element => {
           </HStack>
         </HStack>
         <CampRegistrationLink
-          p={
+          linkUrl={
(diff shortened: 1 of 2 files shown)


frontend/src/components/pages/CampOverview/CampRegistrationLink.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <Button
+        colorScheme="green"
+        borderRadius="0"
+        size="sm"
+        disabled={isDisabled}
+        onClick={() => {
+          navigator.clipboard.writeText(p);
+          toast({
+            title: "Copied to clipboard.",
+            status: "success",
```
dhmi19, line 38: can you make this `variant: subtle` like the other toasts in the code base?

frontend/src/components/pages/CampOverview/CampRegistrationLink.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <InputLeftElement>
+          <LinkIcon color="gray.300" />
+        </InputLeftElement>
+        <Input disabled placeholder={p} borderRadius="0" />
+      </InputGroup>
+      <Button
+        colorScheme="green"
+        borderRadius="0"
+        size="sm"
+        disabled={isDisabled}
```
dhmi19, line 33: I think the designers wanted the button to be in a "copied" and disabled state temporarily for 3 seconds and then change back to the copy CTA. 

We can handle this by passing in another flag saying isCopied which is first set to false but onclick gets set to true. then we can trigger a 3 second timeout after which the state for that flag is set to copied again. Here are some discussions which might be helpful:
https://upmostly.com/tutorials/settimeout-in-react-components-using-hooks

frontend/src/components/pages/CampOverview/CampDetails.tsx
```diff
@@ -110,6 +111,12 @@ const CampDetails = ({ camp, setCamp }: CampDetailsProps): JSX.Element => {
             </Tag>
           </HStack>
         </HStack>
+        <CampRegistrationLink
+          p={
```
dhmi19, line 115: can we rename this prop to something more explicit like `linkURL`?

dhmi19 (changes requested):
forgot to request changes above ^^

dhmi19:
Not sure if there is an issue with my setup but it seems like the "copy" button is always disabled:
image

PR #127, opened 2022-10-08
Create bulk update camp sessions endpoint

e-wai:
## Notion ticket link

[Camp Overview: Delete Session + Change Session Capacity ](https://www.notion.so/uwblueprintexecs/Camp-Overview-Delete-Session-Change-Session-Capacity-6d18cb517d61447ab691f8a7d37978f6)

## Implementation description
* Create bulk update at `PATCH /camp/:camp_id/session`
* Change update endpoint at `PATCH /camp/:camp_id/session/:session_id` to not make capacity and dates both mandatory

## Steps to test

`PATCH /camp/63538bc01bf0453b8a55ea1a/session`

- tested malformed inputs (invalid primitives, non-array body)
- tested happy path, all / partial fields (example body has both)
- tested invalid id, id does not exist cases

```json
{
  "updatedCampSessions": [
    {
      "id": "6353a97ff9b07c73611423ed",
      "capacity": 22
    },
    {
      "id": "6353b23dbd2327c976b7ee6a",
      "capacity": 23, 
      "dates": [
        "Mon Mar 14 2022 12:00:00 GMT+0000 (Coordinated Universal Time)",
        "Wed Mar 16 2022 12:00:00 GMT+0000 (Coordinated Universal Time)",
        "Fri Mar 18 2022 12:00:00 GMT+0000 (Coordinated Universal Time)"
      ]
    }
  ]
}
```

`PATCH /camp/63538bc01bf0453b8a55ea1a/session/6353a97ff9b07c73611423ed`

- test partial bodies, eg.
```json
{
  "capacity": 25,
}
```

* Check use of transactions
* Check optional field logic, does this make sense from a product standpoint?
* There's no objectId validation in campValidator? Do we want it (cast will fail at some point otherwise anyways, but a worse pattern)?

## Checklist

### DIFF
--- a/backend/typescript/middlewares/validators/campValidators.ts
+++ b/backend/typescript/middlewares/validators/campValidators.ts
@@ -342,13 +342,19 @@ export const updateCampSessionDtoValidator = async (
   next: NextFunction,
 ) => {
   const campSession = req.body;
+  if (!campSession.dates && !campSession.capacity) {
+    return res.status(400).send("no fields to update included in request body");
+  }
   if (campSession.dates && !validateArray(campSession.dates, "string")) {
     return res.status(400).send(getApiValidationError("dates", "string", true));
   }
   if (campSession.dates && !campSession.dates.every(validateDate)) {
     return res.status(400).send(getApiValidationError("dates", "Date string"));
   }
-  if (!validatePrimitive(campSession.capacity, "integer")) {
+  if (
+    campSession.capacity &&
+    !validatePrimitive(campSession.capacity, "integer")
+  ) {
     return res.status(400).send(getApiValidationError("capacity", "integer"));
   }
   if (req.body.campers) {
--- a/backend/typescript/middlewares/validators/campValidators.ts
+++ b/backend/typescript/middlewares/validators/campValidators.ts
@@ -387,37 +387,22 @@ export const updateCampSessionsDtoValidator = async (
   res: Response,
   next: NextFunction,
 ) => {
-  if (req.body.campers) {
-    return res
-      .status(400)
-      .send("cannot directly edit campers, field should be empty");
-  }
-  if (req.body.waitlist) {
-    return res
-      .status(400)
-      .send("cannot directly edit waitlist, field should be empty");
-  }
-
-  const { campSessionIds } = req.body;
-  if (!campSessionIds) {
-    return res.status(400).send("campSessionIds are required");
-  }
-
-  for (let index = 0; index < campSessionIds.length; index += 1) {
-    const id = campSessionIds[index];
-    if (!validatePrimitive(id, "string")) {
-      return res.status(400).send(getApiValidationError("id", "string"));
-    }
-  }
-
   const campSessions = req.body.updatedCampSessions;
 
   if (!campSessions) {
-    return res.status(400).send("campSessions are required");
+    return res.status(400).send("updatedCampSessions is a required field");
   }
 
   for (let index = 0; index < campSessions.length; index += 1) {
     const campSession = campSessions[index];
+    if (!campSession.id) {
+      return res
+        .status(400)
+        .send("id is a required field for all camp sessions");
+    }
+    if (!validatePrimitive(campSession.id, "string")) {
+      return res.status(400).send(getApiValidationError("id", "string"));
+    }
     if (campSession.dates && !validateArray(campSession.dates, "string")) {
       return res
         .status(400)
@@ -430,6 +415,16 @@ export const updateCampSessionsDtoValidator = async (
     }
     if (!validatePrimitive(campSession.capacity, "integer")) {
       return res.status(400).send(getApiValidationError("capacity", "integer"));
+    }
+    if (campSession.campers) {
+      return res
+        .status(400)
+        .send("cannot directly edit campers, remove field on request object");
+    }
+    if (campSession.waitlist) {
+      return res
+        .status(400)
+        .send("cannot directly edit waitlist, remove field on request object");
     }
   }
   return next();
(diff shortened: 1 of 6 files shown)


backend/typescript/middlewares/validators/campValidators.ts
```diff
@@ -379,6 +386,56 @@ export const deleteCampSessionsDtoValidator = async (
...
+  res: Response,
+  next: NextFunction,
+) => {
+  const campSessions = req.body.updatedCampSessions;
+
+  if (!campSessions) {
+    return res.status(400).send("updatedCampSessions is a required field");
+  }
+
+  for (let index = 0; index < campSessions.length; index += 1) {
```
dhmi19, line 402: any reason not to use the inbuilt `forEach()` function?

backend/typescript/middlewares/validators/campValidators.ts
```diff
@@ -379,6 +386,56 @@ export const deleteCampSessionsDtoValidator = async (
...
 };
 
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable-next-line import/prefer-default-export */
+export const updateCampSessionsDtoValidator = async (
+  req: Request,
+  res: Response,
+  next: NextFunction,
+) => {
+  const campSessions = req.body.updatedCampSessions;
```
dhmi19, line 396: Is it possible to enforce a CampSession type here?

backend/typescript/middlewares/validators/campValidators.ts
```diff
@@ -379,6 +386,56 @@ export const deleteCampSessionsDtoValidator = async (
...
+    }
+    if (campSession.dates && !campSession.dates.every(validateDate)) {
+      return res
+        .status(400)
+        .send(getApiValidationError("dates", "Date string"));
+    }
+    if (!validatePrimitive(campSession.capacity, "integer")) {
+      return res.status(400).send(getApiValidationError("capacity", "integer"));
+    }
+    if (campSession.campers) {
```
dhmi19, line 425: Can we not make this an `UpdateCampSession` type which the body has to conform to? That way, we can never get campers and waitlist in the body (refer to the typing comment above). Might have missed something here though

dhmi19: oh, i just saw that you do have that type. can we not just enforce the body to conform to that type above? Maybe that's the point of the validator but otherwise, we would have to check for arbitrary fields right?

backend/typescript/services/implementations/campService.ts
```diff
@@ -791,6 +801,117 @@ class CampService implements ICampService {
...
+              session.id,
+            );
+            if (oldCampSession) {
+              if (
+                session.capacity &&
+                oldCampSession.campers &&
+                session.capacity < oldCampSession.campers.length
+              ) {
+                throw new Error(
+                  `Cannot decrease capacity to current number of registered campers. Requested capacity change: ${session.capacity}, current number of registed campers: ${oldCampSession.campers.length}`,
```
dhmi19, line 843: shouldn't this be "cannot decrease capacity to less than current number of registered campers"?

backend/typescript/services/implementations/campService.ts
```diff
@@ -791,6 +801,117 @@ class CampService implements ICampService {
...
+            {
+              ...(session.capacity && { capacity: session.capacity }),
+              ...(session.dates && {
+                dates: session.dates.sort(
+                  (dateA, dateB) =>
+                    new Date(dateA).getTime() - new Date(dateB).getTime(),
+                ),
+              }),
+            },
+            { runValidators: true, new: true },
```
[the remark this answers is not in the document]
dhmi19: I suspect it makes a new document if `findByIdAndUpdate` doesn't find a document with the given `session.id`. But this should never be the case no?

dhmi19 (commented):
Left a few comments, mostly asking for clarification on some things. Looks good otherwise

backend/typescript/middlewares/validators/campValidators.ts
e-wai: Hmm we haven't brought types into this file before, didn't really wanna mess with that pattern?

backend/typescript/middlewares/validators/campValidators.ts
e-wai: iirc there was the `forEach` linter problems, maybe https://github.com/github/eslint-plugin-github/issues/86?

backend/typescript/middlewares/validators/campValidators.ts
e-wai: Yeahh it's not great, I'm wondering if I'm missing some sort of `try { <cast> } catch { return res.status(400) }` pattern? I only check for these because they're fields on the object that may lead to common misuse

backend/typescript/services/implementations/campService.ts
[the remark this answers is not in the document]
e-wai: `[options.new=false] «Boolean» if true, return the modified document rather than the original`

https://mongoosejs.com/docs/api.html#model_Model-findByIdAndUpdate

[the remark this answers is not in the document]
dhmi19: thanks, looks good

backend/typescript/middlewares/validators/campValidators.ts
dhmi19: ok, no big deal

backend/typescript/middlewares/validators/campValidators.ts
dhmi19: fair enough

dhmi19 (approved):
addressed all my comments. looks good to go

PR #132, opened 2022-10-15
Waiver card view only

jasminekbal:
## Notion ticket link

[Waiver Card View Only](https://www.notion.so/uwblueprintexecs/Form-Management-Waiver-Waiver-Card-View-Only-Page-Layout-3690a9e4019b4acf8548eb2f63c0721a)

## Implementation description
*  I created a new folder, GlobalForms, for the FormManagement Page and added two new components to the folder. The first is the card for the waiver, the second is a wrapper around the card that iterates through all Waivers.
* Also added an AdminAPIClient file to call the getWaiver() function for the /admin/waiver end point. An AdminTypes file was also added.

## Steps to test
1. Check that the cards are shown properly and that if data is changed in MongoDB, it is reflected in the cards.

* 

## Checklist
- [x ] My PR name is descriptive and in imperative tense
- [ x] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [x ] I have run the appropriate linter(s)
- [ x] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR


frontend/src/APIClients/AdminAPIClient.ts
```diff
@@ -0,0 +1,17 @@
...
+import baseAPIClient from "./BaseAPIClient";
+
+const getWaiver = async () => {
+  try {
+    const { data } = await baseAPIClient.get(`/admin/waiver`, {
+      headers: { Authorization: BEARER_TOKEN },
+    });
+    return data;
+  } catch (error) {
+    return {};
```
dhmi19, line 11: Can you return the following instead:
```
catch(error) {
    return error as WaiverResponse;
}
```
I am not sure how the parsing is working but looking at other API clients, we return the error casted as the return type when we get an error so it should work

frontend/src/components/pages/GlobalForms/GlobalFormsPage.tsx
```diff
@@ -0,0 +1,34 @@
...
+import AdminAPIClient from "../../../APIClients/AdminAPIClient";
+import { ClauseInfo } from "../../../types/AdminTypes";
+import WaiversTab from "./WaiversTab";
+
+const GlobalFormsPage = (): React.ReactElement => {
+
+  const [waiver, setWaiver] = React.useState([]as ClauseInfo[]);
+  React.useEffect(() => {
+    const getWaiver = async () => {
+      const res = await AdminAPIClient.getWaiver();
```
dhmi19, line 12: If you define the return type in the getWaiver function, then you'll know the type of the res variable here. That way you can check if you got a proper waiver back. A good example is on the camp overview `index.ts` component which makes a get request to retrieve all the info for that particular camp:

```
useEffect(() => {
    const getCamp = async () => {
      const campResponse = await CampsAPIClient.getCampById(campId);
      if (campResponse) {
        setCamp(campResponse);
      }
    };
    getCamp();
  }, [campId]);
```

frontend/src/components/pages/GlobalForms/GlobalFormsPage.tsx
```diff
@@ -0,0 +1,34 @@
...
+      if (res.clauses.length !== undefined) setWaiver(res.clauses);
+    };
+
+    getWaiver();
+  }, []);
+
+  return (
+    <Container
```
dhmi19, line 20: Can you run the following 2 commands to make sure you lint your code before merging? Linting is a way to auto-format our code based on the agreed formats we set up at the start of the project. Run these 2 commands whilst you have your Docker containers running:

`docker exec -it scv2_ts_backend /bin/bash -c "yarn fix"`

`docker exec -it scv2_frontend /bin/bash -c "yarn fix"`

frontend/src/components/pages/GlobalForms/GlobalFormsPage.tsx
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+import {Text, Container} from '@chakra-ui/react'
+import AdminAPIClient from "../../../APIClients/AdminAPIClient";
+import { ClauseInfo } from "../../../types/AdminTypes";
+import WaiversTab from "./WaiversTab";
+
+const GlobalFormsPage = (): React.ReactElement => {
+
+  const [waiver, setWaiver] = React.useState([]as ClauseInfo[]);
```
dhmi19, line 9: If you're using waiver as the state, I would recommend making this state of the type `WaiverResponse` 
This would look like:
```
const [waiver, setWaiver] = React.useState<WaiverResponse>()
```

Then instead of setting the state to the set of clauses, you would just set the state to whatever your APIClient getWaiver function returns since you are enforcing the return type of that function to `WaiverResponse`

dhmi19: If you want to use the clauses only, I would suggest renaming the state to waiverClauses so it's a bit more clear

frontend/src/components/pages/GlobalForms/WaiversTab.tsx
```diff
@@ -0,0 +1,28 @@
...
+    clauses
+  }: WaiversTabProps): JSX.Element => {
+
+    function intToChar(number: number) {
+        const code = 'A'.charCodeAt(0);  
+        return String.fromCharCode(code + number);
+    }
+    return (
+        <>
+        {clauses.length && (clauses).map((clause, idx)=> {
```
dhmi19, line 18: You should be able to map over an empty array. So you should be able to replace this with:
```
{
    clauses.map((clause, index) => {..})
}
```

frontend/src/components/pages/GlobalForms/WaiversTab.tsx
```diff
@@ -0,0 +1,28 @@
...
+            return <WaiverSectionCard key={idx} title={`Section ${intToChar(idx)}`}
```
dhmi19, line 19: Can you rename the key to  " (backtick) waiver_section_card_${index} (backtick) ". this makes the key more meaningful instead of just a number. 

Note (backtick) represents a ` . Github kept changing that to think of the text as code

frontend/src/components/pages/GlobalForms/WaiversTab.tsx
```diff
@@ -0,0 +1,28 @@
...
+            waiverText={clause.text}
```
dhmi19, line 20: Instead of passing in each property of the clause, can you just pass in the whole clause? This would just make code a little cleaner

dhmi19: so you only need to pass in the clause and the index to the `WaiverSectionCard` component. The rest can be handled inside the component itself

frontend/src/components/pages/GlobalForms/WaiversTab.tsx
```diff
@@ -0,0 +1,28 @@
...
+import WaiverSectionCard from "./WaiverSectionCard";
+  
+  interface WaiversTabProps {
+    clauses: Array<ClauseInfo>;
+  }
+  const WaiversTab = ({
+    clauses
+  }: WaiversTabProps): JSX.Element => {
+
+    function intToChar(number: number) {
```
dhmi19, line 12: I think we can move this function to the Card component. Can we also rename this function something like `getSectionTitle` which takes an index and just returns the full title? (eg: takes index = 1 and returns "Section B")

This would make it easier to read the code for others instead of having to understand how you're converting the index to character

frontend/src/components/pages/GlobalForms/WaiverSectionCard.tsx
```diff
@@ -0,0 +1,35 @@
+import {
+    Text, Box, Flex, HStack, Spacer,
+  } from "@chakra-ui/react";
+  import React from "react";
```
dhmi19, line 4: Can you move the React import to the top of the file?

dhmi19 (changes requested):
This looks great! I left some minor feedback but the general idea is solid

dhmi19 (changes requested):
looks good, thanks for making the changes so fast! I think you just missed one comment which I unresolved. We'll merge in this branch after we merge in Jasraj and Ryan's PRs

dhmi19 (approved):

PR #137, opened 2022-10-23
chore: clean up eslint warnings

hanlinc27:
## Notion ticket link
no ticket, contracted work

## Implementation description
fix some types, delete some stuff

## Steps to test
I have no secrets so...make sure the FE server runs..?

that nothing is broken

## Checklist
- [X] My PR name is descriptive and in imperative tense
- [X] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [X] I have run the appropriate linter(s)
- [X] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR


frontend/src/components/pages/AccessControl/index.tsx
```diff
@@ -90,7 +90,13 @@ const AccessControlPage = (): JSX.Element => {
           .concat(" ", user.lastName.toLowerCase())
           .includes(search.toLowerCase()),
     );
-  }, [search, selectedFilter, users]);
+  }, [
+    search,
+    selectedFilter,
+    users,
+    UserStatusStates.ACTIVE_CAPITALIZED,
+    UserStatusStates.ALL,
```
hanlinc27, line 98: the other option is to just transform this into a callback but throwing these into the dep array should have no effect since the enums aren't being mutated

frontend/src/APIClients/AuthAPIClient.ts
```diff
@@ -23,9 +23,7 @@ const login = async (
   }
 };
 
-const loginWithGoogle = async (
-  idToken: string,
-): Promise<AuthenticatedUser | string> => {
+const loginWithGoogle = async (idToken: string): Promise<AuthenticatedUser> => {
```
hanlinc27, line 26: this is probably the only change I'm a bit iffy on, was there a reason why you were returning a string as a type _ever_?

frontend/src/components/auth/Login.tsx
```diff
@@ -51,7 +45,7 @@ const Login = (): React.ReactElement => {
       tokenId,
     );
 
-    if (isAuthenticatedUserType(res)) {
+    if (res) {
```
hanlinc27, line 48: also think we can just check `res` exists here instead of having the whole function extracted right?

frontend/src/components/pages/CampOverview/index.tsx
```diff
@@ -116,7 +116,9 @@ const CampOverviewPage = (): JSX.Element => {
                   };
                 })}
               onSaveChanges={(deletedSessions, updatedCapacities) => {
+                /* eslint-disable no-console */
```
hanlinc27, line 119: hacky but no real better solution, I think a `window.alert` would probably also throw a warning

e-wai (approved):
Ty 😍

PR #138, opened 2022-10-26
added the addQuestionToFormTemplate in the admin service

dhmi19:
## Notion ticket link

[Form Management: Add question to form template](https://www.notion.so/uwblueprintexecs/Task-Board-F22-9e03bf9f7c4343f09a8a5c16dc48a12a?p=872084408d3543a3ad120dcc9808a9a1&pm=c)

## Implementation description
* just your standard endpoint changes tbh
* decided to make the body of the form `formQuestion: {...}` instead of just the fields. This was to leverage the existing `validateFormQuestion()` function more directly. 
* Decided to implement this function in the adminService as I think it does belong there. Some of the existing work for formquestions seems to be in the `campService` but I think it's better suited here. (For camp creation, I think we should add new functions like `addQuestionsToCamp` like the one I added in this PR)

## Steps to test
1. Make a PATCH request to `http://localhost:5000/admin/formTemplate/formQuestion/` (you might need to disable the `isAuthorizedByRole()` middleware or figure out how to add auth to POSTMAN requests)
2. Use the following body: 
```
{
    "formQuestion" : {
        "type" : "Text",
        "category": "PersonalInfo",
        "question": "What previous photography experience do you have?",
        "required" : false
    }
}
```
3. Log on to the mongodb console and see that the question id was APPENDED to the list of questions in the template and that the question was created in the `formQuestion` collection
4. The response should have the newly created formQuestion

* Use of transactions ok? 

## Checklist

### DIFF
--- a/backend/typescript/middlewares/validators/adminValidators.ts
+++ b/backend/typescript/middlewares/validators/adminValidators.ts
@@ -49,3 +49,14 @@ export const formTemplateUpdateValidator = async (
   }
   return next();
 };
+
+export const formTemplateAddQuestionValidator = async (
+  req: Request,
+  res: Response,
+  next: NextFunction,
+) => {
+  if (!req.body.formQuestion || !validateFormQuestion(req.body.formQuestion)) {
+    return res.status(400).send("Invalid format provided for form question");
+  }
+  return next();
+};
--- a/backend/typescript/rest/adminRoutes.ts
+++ b/backend/typescript/rest/adminRoutes.ts
@@ -6,10 +6,13 @@ import { getErrorMessage } from "../utilities/errorUtils";
 import {
   waiverUpdateValidator,
   formTemplateUpdateValidator,
+  formTemplateAddQuestionValidator,
 } from "../middlewares/validators/adminValidators";
+import { isAuthorizedByRole } from "../middlewares/auth";
 
 const adminRouter: Router = Router();
 const adminService: IAdminService = new AdminService();
+
 adminRouter.post("/waiver", waiverUpdateValidator, async (req, res) => {
   try {
     const waiver = await adminService.updateWaiver({
@@ -53,5 +56,26 @@ adminRouter.get("/formTemplate", async (req, res) => {
     res.status(500).json({ error: getErrorMessage(error) });
   }
 });
+
+adminRouter.patch(
+  "/formTemplate/formQuestion",
+  formTemplateAddQuestionValidator,
+  // isAuthorizedByRole(new Set(["Admin"])),
+  async (req, res) => {
+    try {
+      const newQuestion = await adminService.addQuestionToTemplate({
+        question: req.body.formQuestion.question,
+        required: req.body.formQuestion.required,
+        category: req.body.formQuestion.category,
+        options: req.body.formQuestion.options,
+        description: req.body.formQuestion.description,
+        type: req.body.formQuestion.type,
+      });
+      res.status(200).json(newQuestion);
+    } catch (error: unknown) {
+      res.status(500).json({ error: getErrorMessage(error) });
+    }
+  },
+);
 
 export default adminRouter;
--- a/backend/typescript/rest/adminRoutes.ts
+++ b/backend/typescript/rest/adminRoutes.ts
@@ -60,7 +60,7 @@ adminRouter.get("/formTemplate", async (req, res) => {
 adminRouter.patch(
   "/formTemplate/formQuestion",
   formTemplateAddQuestionValidator,
-  // isAuthorizedByRole(new Set(["Admin"])),
+  isAuthorizedByRole(new Set(["Admin"])),
   async (req, res) => {
     try {
       const newQuestion = await adminService.addQuestionToTemplate({
--- a/backend/typescript/services/implementations/adminService.ts
+++ b/backend/typescript/services/implementations/adminService.ts
@@ -123,11 +123,11 @@ class AdminService implements IAdminService {
 
     try {
       // Get the current formTemplate and its list of questions
-      const formTemplate = await MgFormTemplate.findOne();
-      
-      if (!formTemplate){
+      const formTemplate = await MgFormTemplate.findOne({}, {}, { session });
+
+      if (!formTemplate) {
         throw new Error("Failed to retrieve the form template");
-      }  
+      }
 
       const newFormQuestion = await MgFormQuestion.create({
         type: formQuestion.type,
@@ -139,7 +139,7 @@ class AdminService implements IAdminService {
       });
 
       formTemplate.formQuestions.push(newFormQuestion._id);
-      await formTemplate.save();
+      await formTemplate.save({ session });
       await session.commitTransaction();
 
       return {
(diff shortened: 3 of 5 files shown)


dhmi19:
### TEST CASES

## Case 1: Happy path
body : 
```
{
    "formQuestion" : {
        "type" : "Text",
        "category": "PersonalInfo",
        "question": "What previous photography experience do you have?",
        "required" : false
    }
}
```
response:
```
{
    "type": "Text",
    "question": "What previous photography experience do you have?",
    "required": false,
    "options": [],
    "id": "6358ae8d06d07d99bcf5d383",
    "category": "PersonalInfo"
}
```

##Case 2: Unauthorized
Same body as above but this time had the middleware uncommented and no authorization on the POSTMAN request
response:
```
{
    "error": "You are not authorized to make this request."
}
```

##Case 3: Bad body (missing the field `required`) 
body:
```
{
    "formQuestion" : {
        "type" : "Text",
        "category": "PersonalInfo",
        "question": "What previous photography experience do you have?"
    }
}
```
response: `Invalid format provided for form question`

backend/typescript/services/implementations/adminService.ts
```diff
@@ -108,6 +114,52 @@ class AdminService implements IAdminService {
...
+        category: newFormQuestion.category,
+      };
+    } catch (error: unknown) {
+      Logger.error(
+        `Failed to add form question to form template. Reason = ${getErrorMessage(
+          error,
+        )}`,
+      );
+      await session.abortTransaction();
+      await session.endSession();
```
e-wai, line 159: should this be moved to a `finally {}`? Even if no error, we should end session

backend/typescript/services/implementations/adminService.ts
```diff
@@ -108,6 +114,52 @@ class AdminService implements IAdminService {
...
+    formQuestion: CreateFormQuestionDTO,
+  ): Promise<FormQuestionDTO> {
+    const session = await mongoose.startSession();
+    session.startTransaction();
+
+    try {
+      // Get the current formTemplate and its list of questions
+      const formTemplate = await MgFormTemplate.findOne();
+      if (!formTemplate)
+        throw new Error("Failed to retrieve the form template");
```
e-wai, line 128: nit: brackets

backend/typescript/services/implementations/adminService.ts
```diff
@@ -108,6 +114,52 @@ class AdminService implements IAdminService {
...
+      const newFormQuestion = await MgFormQuestion.create({
+        type: formQuestion.type,
+        question: formQuestion.question,
+        required: formQuestion.required,
+        description: formQuestion.description,
+        options: formQuestion.options,
+        category: formQuestion.category,
+      });
+
+      formTemplate.formQuestions.push(newFormQuestion._id);
```
e-wai, line 139: I've completely forgotten, what's the difference between `_id` and `id`?

e-wai:
> Make a PATCH request to http://localhost:5000/admin/formTemplate/formQuestion/ (you might need to disable the isAuthorizedByRole() middleware or figure out how to add auth to POSTMAN requests)

I think this restriction is figured out by your auth token so should be good in POSTMAN

backend/typescript/middlewares/validators/adminValidators.ts
```diff
@@ -49,3 +49,14 @@ export const formTemplateUpdateValidator = async (
...
   return next();
 };
+
+export const formTemplateAddQuestionValidator = async (
+  req: Request,
+  res: Response,
+  next: NextFunction,
+) => {
+  if (!req.body.formQuestion || !validateFormQuestion(req.body.formQuestion)) {
+    return res.status(400).send("Invalid format provided for form question");
```
e-wai, line 59: I think it's worth specifying which field we're missing?

e-wai (changes requested):
Transactions 😍 so much cleaner, also tysm for the documentation on the testing cases, works like a charm

> Decided to implement this function in the adminService as I think it does belong there. Some of the existing work for formquestions seems to be in the campService but I think it's better suited here. (For camp creation, I think we should add new functions like addQuestionsToCamp like the one I added in this PR)

yep agree

- Should this be a POST endpoint? It always creates new resources
- I also didn't realize none of our endpoints are protected... they very much should be

backend/typescript/rest/adminRoutes.ts
```diff
@@ -54,4 +57,25 @@ adminRouter.get("/formTemplate", async (req, res) => {
   }
 });
 
+adminRouter.patch(
+  "/formTemplate/formQuestion",
+  formTemplateAddQuestionValidator,
+  // isAuthorizedByRole(new Set(["Admin"])),
```
e-wai, line 63: Could we uncomment this (to test, I can just change my permissions)

backend/typescript/services/implementations/adminService.ts
dhmi19: the id is the string representation and _id is the actual OjectId type

dhmi19:
> Transactions 😍 so much cleaner, also tysm for the documentation on the testing cases, works like a charm
> 
> > Decided to implement this function in the adminService as I think it does belong there. Some of the existing work for formquestions seems to be in the campService but I think it's better suited here. (For camp creation, I think we should add new functions like addQuestionsToCamp like the one I added in this PR)
> 
> yep agree
> 
> * Should this be a POST endpoint? It always creates new resources
> * I also didn't realize none of our endpoints are protected... they very much should be

I was considering POST but we are technically updating the form template. I don't think we can ever create a form question by itself (the creation will always either be part of updating the template or the camp's list of form questions)

backend/typescript/services/implementations/adminService.ts
dhmi19: good callout

e-wai (approved):
LGTM, just linter outstanding

backend/typescript/services/implementations/adminService.ts
```diff
@@ -108,6 +114,55 @@ class AdminService implements IAdminService {
     }
     return form;
   }
+
+  async addQuestionToTemplate(
+    formQuestion: CreateFormQuestionDTO,
+  ): Promise<FormQuestionDTO> {
+    const session = await mongoose.startSession();
+    session.startTransaction();
```
rchang0501, line 122: ah finally understand what a transaction is - cool!

backend/typescript/services/interfaces/adminService.ts
```diff
@@ -8,24 +13,36 @@ interface IAdminService {
...
    * Get the form.
    * @returns A FormTemplateDTO that contains the waiver.
    * @throws Error if form template retrieval fails.
    */
   getFormTemplate(): Promise<FormTemplateDTO>;
+
+  /**
+   * Add a question to the form template
+   * @returns a FormQuestionDTO for the new form
+   * @throws Error if adding the form question fails
```
rchang0501, line 41: should we also define params? I know we use them in the camper service interface, how come we don't use them here?

rchang0501 (approved):
lgtm, just small comment for my own curiosity

dhmi19:
TODO: forgot to include {session} in my mongoose db calls

backend/typescript/services/interfaces/adminService.ts
dhmi19: good catch. I forgot about those

e-wai:
> TODO: forgot to include {session} in my mongoose db calls

Like as an import? How did it work without the import?

backend/typescript/middlewares/validators/adminValidators.ts
e-wai: Yeah, fair enough

PR #143, opened 2022-10-31
Frontend - Waiver Section Edit/Delete

jasminekbal:
## Notion ticket link

[Edit + Delete Front end ](https://www.notion.so/uwblueprintexecs/Task-Board-F22-9e03bf9f7c4343f09a8a5c16dc48a12a?p=b4dbb43b89144b8f8b4d54ba04725a0e&pm=c)

## Implementation description
* I added two modals for the edit and delete confirmation in the WaiverSectionCard.ts file
* I added two functions onDeleteWaiverSection and onEditWaiverSection in the GlobalForms/index.ts file and passed these down so that we could call the API at the highest level instead of changing state multiple times

## Steps to test
1. Try editing a section -> the section should change and there should be a success toast
2. Try deleting a section -> the section should disappear and there should be a success toast

* I'm not sure if I'm handling state the best way in the WaiverSectionCard.ts file, I went back and forth with splitting up the modals into multiple files but ended up thinking it would be nice if everything related to one section was in the same file

## Checklist
- [ x] My PR name is descriptive and in imperative tense
- [x ] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [ x] I have run the appropriate linter(s)
- [ x] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR

### DIFF
--- a/frontend/src/components/pages/GlobalForms/WaiverTab/WaiverSectionCard.tsx
+++ b/frontend/src/components/pages/GlobalForms/WaiverTab/WaiverSectionCard.tsx
@@ -1,21 +1,62 @@
 import React from "react";
-import { Text, Box, Flex, HStack, Spacer } from "@chakra-ui/react";
+import { Text, Box, Flex, HStack, Spacer, Modal,
+  ModalOverlay,
+  ModalContent,
+  ModalHeader,
+  ModalFooter,
+  ModalBody,
+ Button, Textarea, Checkbox } from "@chakra-ui/react";
 import { WaiverClause } from "../../../../types/AdminTypes";
 
 interface WaiverSectionCardProps {
   clauseIdx: number;
   clauseData: WaiverClause;
+  onEditWaiverSection: (clauseText: string, clauseIsRequired: boolean, clauseIdx: number) => void;
+  onDeleteWaiverSection: (idx: number) => void;
 }
 const WaiverSectionCard = ({
   clauseIdx,
   clauseData,
+  onEditWaiverSection,
+  onDeleteWaiverSection,
 }: WaiverSectionCardProps): JSX.Element => {
   function getSectionTitle(number: number) {
     const code = "A".charCodeAt(0);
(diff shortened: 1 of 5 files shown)


frontend/src/components/pages/GlobalForms/WaiverTab/WaiverSectionCard.tsx
```diff
@@ -1,50 +1,180 @@
...
+  Textarea,
+  Checkbox,
+} from "@chakra-ui/react";
 import { WaiverClause } from "../../../../types/AdminTypes";
 
 interface WaiverSectionCardProps {
   clauseIdx: number;
   clauseData: WaiverClause;
+  onEditWaiverSection: (
+    clauseText: string,
```
dhmi19, line 24: Instead of listing the fields in the `onEditWaiverSection` function here, could you instead pass in a `WaiverClause` object? The benefit of this is that if we change the definition of the WaiverClause, then we don't need to update the function everywhere with the new fields etc.

frontend/src/components/pages/GlobalForms/WaiverTab/WaiverSectionCard.tsx
```diff
@@ -1,50 +1,180 @@
...
             >
-              required
+              Delete
+            </Text>
+          </HStack>
+        </Flex>
+        <Text>{clauseData.text}</Text>
+      </Box>
+      <>
+        <Modal isOpen={isEditOpen} onClose={onEditClose} isCentered>
```
dhmi19, line 112: Can you add the preserveScrollBarGap prop to the Modals?

dhmi19 (changes requested):
Almost there! Love the functionality and how you preserved the state to update the UI immediately. Left a few minor comments for some changes but excited to merge this asap!

dhmi19 (approved):

PR #145, opened 2022-11-05
Add stepper navigation in camp creation flow

twilkhoo:
## Notion ticket link

[Camp Creation/Editing: Stepper Navigation Component](https://www.notion.so/uwblueprintexecs/Camp-Creation-Editing-Stepper-Navigation-Component-001099c4e1004cc784b3e18fc8be5e09)

## Implementation description
* Created the stepper, implemented as boxes with `onClick`s
* Added dummy state in the main camp creation page component

## Steps to test
1. Go to my branch and try it out locally or see below (paying attention to when I try to go to a following page but am unable to)

https://user-images.githubusercontent.com/30396273/201606273-9bcb618d-2fbf-4eb1-8e59-7105775155aa.mp4

* Review all the design decisions/considerations I made, most of the dummy code only exists in the main component since we don't have anything like Redux right now.

## Checklist


frontend/src/components/pages/CampCreation/index.tsx
```diff
@@ -1,58 +1,95 @@
...
 import ScheduleSessions from "./ScheduleSessions";
 import RegistrationForm from "./RegistrationForm";
+import Stepper, { CampCreationPages } from "./StepperNavigation/index";
 
-enum CampCreationPages {
-  CampCreationDetailsPage,
-  ScheduleSessionsPage,
-  RegistrationFormPage,
-}
+const CampCreationPage = (): React.ReactElement => {
```
twilkhoo, line 8: I think we'll (unfortunately) need to have all response state in this main component, since having it in the subpage components themselves don't persist the state when we navigate between pages.

frontend/src/components/pages/CampCreation/index.tsx
```diff
@@ -1,58 +1,95 @@
...
+  );
+  const [registrationFormDummyTwo, setRegistrationFormDummyTwo] = useState(
+    false,
+  );
+
+  // State to determine whether or not all required fields have been filled out.
+  // NOTE: This will depend on what type of state a page requires, i.e. determining
+  // if a checkbox is checked is different than determining if an input field is filled.
+
+  const isCampDetailsFilled =
```
twilkhoo, line 30: Each page is going to have a different definition of what's required to be filled (some may be optional fields that wouldn't show up here).

frontend/src/components/pages/CampCreation/index.tsx
```diff
@@ -1,58 +1,95 @@
...
+        isScheduleSessionsFilled={isScheduleSessionsFilled}
+        isRegistrationFormFilled={isRegistrationFormFilled}
+      />
+      <Box m={20}>
+        {currentPage === CampCreationPages.CampDetailsPage && (
+          <CampDetails
+            campDetailsDummyOne={campDetailsDummyOne}
+            campDetailsDummyTwo={campDetailsDummyTwo}
+            campDetailsDummyThree={campDetailsDummyThree}
+            toggleCampDetailsDummyOne={() =>
```
twilkhoo, line 56: It seems redundant but we probably need to pass in both getters and setters for each piece of state as props... since these are controlled components, something like a checkbox would need both the getter and setter.

frontend/src/components/pages/CampCreation/StepperNavigation/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  isScheduleSessionsFilled: boolean;
+  isRegistrationFormFilled: boolean;
+};
+
+const Stepper = ({
+  currentPage,
+  setCurrentPage,
+  isCampDetailsFilled,
+  isScheduleSessionsFilled,
+  isRegistrationFormFilled,
```
twilkhoo, line 30: We aren't currently using this prop in the stepper but we probably will need it later (when we try to submit the camp to the db)

frontend/src/components/pages/CampCreation/StepperNavigation/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { Box, Center } from "@chakra-ui/react";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import {
+  faCampground,
+  faCalendarDays,
+  faCircleInfo,
+} from "@fortawesome/free-solid-svg-icons";
+import StepperTab from "./StepperTab";
+
+export enum CampCreationPages {
```
dhmi19, line 11: Not that it matters all that much, but why define the pages here instead of the parent wrapper component?

frontend/src/components/pages/CampCreation/StepperNavigation/index.tsx
dhmi19: would we need this prop in the Stepper component though? I think it would be useful in the footer so we can conditionally render the `publish camp` button or use it for validation

frontend/src/components/pages/CampCreation/StepperNavigation/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const Stepper = ({
+  currentPage,
+  setCurrentPage,
+  isCampDetailsFilled,
+  isScheduleSessionsFilled,
+  isRegistrationFormFilled,
+}: StepperProps): React.ReactElement => {
+  return (
+    <Box bg="white" maxW="100%" boxShadow="sm">
```
dhmi19, line 33: is the `maxW` prop necessary here?

frontend/src/components/pages/CampCreation/StepperNavigation/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+          focused={currentPage === CampCreationPages.CampCreationDetailsPage}
+          icon={<FontAwesomeIcon icon={faCampground} />}
+          onClick={() => setCurrentPage(CampCreationPages.CampCreationDetailsPage)}
+        />
+        <StepperTab
+          title="Schedule Sessions"
+          stepNum={2}
+          focused={currentPage === CampCreationPages.ScheduleSessionsPage}
+          icon={<FontAwesomeIcon icon={faCalendarDays} />}
+          onClick={() =>
```
dhmi19, line 47: this is great but we might want to consider disabling this button if the previous step(s) are not filled. Something to differentiate between future and past steps would be useful imo. Waiting on design suggestions

frontend/src/theme/colors.ts
```diff
@@ -37,7 +37,9 @@ const colors = {
       100: "#FFFFFF",
     },
     grey: {
-      100: "#A3AEBE",
+      50: "#FAFAFA",
+      100: "#FAFAFA",
```
dhmi19, line 41: What's the difference between the 50 and 100?

dhmi19 (changes requested):
Look solid! Honestly, I would say this is pretty much good to merge but I want to get some design feedback since we have the bandwidth to improve the visuals if needed in this sprint

frontend/src/components/pages/CampCreation/index.tsx
dhmi19: Yeah i think we can even abstract this to be a function if needed. Although a constant works well for now

frontend/src/components/pages/CampCreation/index.tsx
dhmi19: I think it's fine given that we don't have something like redux set up. If it gets too ugly, we can do what I did for edit camper info where we would defined a type for each page for setters and gettters and then pass in that const instead of listing all the props.

frontend/src/components/pages/CampCreation/StepperNavigation/StepperTab.tsx
```diff
@@ -0,0 +1,53 @@
...
+              colorScheme="text.grey"
+              border="2px"
+              isRound
+              size="lg"
+              aria-label={title}
+            />
+          )}
+        </Box>
+        <Box>
+          <Text textStyle="caption">Step {stepNum}/3</Text>
```
e-wai, line 45: can we make `totalSteps` a parameter? It can be used more generically for the 4 steps in the registration process

frontend/src/components/pages/CampCreation/StepperNavigation/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export enum CampCreationPages {
+  CampCreationDetailsPage,
+  ScheduleSessionsPage,
+  RegistrationFormPage,
+}
+
+type StepperProps = {
+  currentPage: CampCreationPages;
+  setCurrentPage: React.Dispatch<React.SetStateAction<CampCreationPages>>;
```
e-wai, line 19: What's the difference between this and `(campCreationPage: CampCreationPage) => void`?

frontend/src/components/pages/CampCreation/index.tsx
```diff
@@ -1,58 +1,95 @@
...
+    <>
+      <Stepper
+        currentPage={currentPage}
+        setCurrentPage={setCurrentPage}
+        isCampDetailsFilled={isCampDetailsFilled}
+        isScheduleSessionsFilled={isScheduleSessionsFilled}
+        isRegistrationFormFilled={isRegistrationFormFilled}
+      />
+      <Box m={20}>
+        {currentPage === CampCreationPages.CampCreationDetailsPage && (
```
e-wai, line 51: Think it's worth extracting this logic into a switch case and its own function?

e-wai (changes requested):
Will do a more thorough lookthrough tomorrow -- biggest suggestion right now is that the `StepperNavigation` component belongs in the "common" components folder instead, since we will eventually be using it in registrant experience

frontend/src/components/pages/CampCreation/StepperNavigation/index.tsx
twilkhoo: Removed it.

frontend/src/components/pages/CampCreation/StepperNavigation/index.tsx
twilkhoo: By parent wrapper component do you mean the Camp Creation index.tsx? If so, we have some circular dependencies since I'll need to import `CampCreationPages` from the camp creation index, but that index will import this Stepper index.

e-wai:
@twilkhoo should I fix this up?

cc: @dhmi19

twilkhoo:
> @twilkhoo should I fix this up?
> 
> cc: @dhmi19

I'm adding some of the display changes right now, I'll message you once those are in.

frontend/src/components/pages/CampCreation/StepperNavigation/index.tsx
dhmi19: ahh ok yeah that makes sense. thanks!

e-wai (approved):
Need to clarify why the colour changes are needed, will merge after that

PR #154, opened 2022-11-17
Add dedicated colorScheme for stepper tab focused state

twilkhoo:
## Implementation description
* Added a colorScheme for the stepper tab focused state

## Steps to test
1. Look:

Before:
image
After:
image

* -

## Checklist


dhmi19 (approved):
lgtm. Emma might want this to be done inline, but I think since we're gonna reuse these colours on the registrant part, it's worth having here. + we define some shared theme in the central themes file too

PR #162, opened 2022-11-23
feat: edit delete form question

williamtran10:
## Notion ticket link

[Form Management: Registration Form Template - Edit/Delete Question (Frontend)](https://www.notion.so/uwblueprintexecs/Form-Management-Registration-Form-Template-Edit-Delete-Question-Frontend-d6ae1b27819640bc869c0499b0e45098)

## Implementation description
* Changed BE functions to return form question id
* Added functions in AdminAPIClient to edit/delete questions
* Added calls to AdminAPIClient
* Used Tejas' add/delete question modals

## Steps to test
1. Go to form question page
2. Click edit/delete

* I switched a lot of `FormQuestion` to `CreateFormQuestion` so that I can utilize the `id` attribute
* I changed the type of the formQuestions from a tuple to an array to match how it was done in other places in the code (plus its easier to work with)
* I made the backend code explicitly return the formQuestion's `id`, which is more verbose but otherwise it would only have an `_id` and not an `id`

## Checklist


frontend/src/components/pages/GlobalForms/index.tsx
```diff
@@ -194,6 +194,45 @@ const GlobalFormsPage = (): React.ReactElement => {
...
+      const newArr = [...oldArr];
+      for (let i = 0; i < newArr.length; i += 1) {
+        if (newArr[i] === oldQuestion) {
+          newArr[i] = newQuestion;
+          break;
+        }
+      }
+      return newArr;
+    });
+
```
dhmi19, line 226: I think you missed the API call to the backend to edit and delete the form questions from the template. 
See the following endpoints:
1. remove question from form template `DELETE "/formTemplate/formQuestion/:formQuestionId"`
2. edit question in template `PATCH "/formTemplate/formQuestion/:oldQuestionId"`

dhmi19 (changes requested):
I think you missed making the calls to the backend endpoints to actually edit and remove questions in the form template. I also merged in Tejas's changes so could you rebase off dev?

frontend/src/components/pages/GlobalForms/index.tsx
williamtran10: Added now

backend/typescript/services/implementations/adminService.ts
```diff
@@ -112,7 +112,19 @@ class AdminService implements IAdminService {
       );
       throw error;
     }
-    return form;
+    return {
+      formQuestions: form.formQuestions.map((formQuestion: FormQuestionDTO) => {
```
dhmi19, line 116: why could we not have returned the `form` here?

dhmi19: I am guessing it's because of the | null. In other functions we rely on returning in the try block and then the catch block would return an error. Would that work here?

frontend/src/components/pages/GlobalForms/index.tsx
```diff
@@ -194,6 +194,67 @@ const GlobalFormsPage = (): React.ReactElement => {
     }
   };
 
+  const onDeleteCustomQuestion = async (formQuestion: FormQuestion) => {
+    const res = await AdminAPIClient.deleteFormQuestion(formQuestion.id);
```
dhmi19, line 198: can we call this function `removeFormQuestionFromTemplate`?

dhmi19 (approved):
Functionality looks good, just had a couple of questions on the code

backend/typescript/services/implementations/adminService.ts
williamtran10: I found experimentally that when `form` was returned, the `FormQuestion`s would only have an `_id` and not an `id` associated with them.

dhmi19: oh right I missed that. I see that we are using a similar approach in other functions too

PR #168, opened 2022-11-28
fix: Replace BEARER_TOKEN constant to be a function

AmandaMisjuwar:
## Notion ticket link

[Access management table not loading on first visit](https://www.notion.so/uwblueprintexecs/Access-Management-Table-not-loading-on-first-visit-0ccaa7ab18504bb787071999442079c9)

## Implementation description
* The access management table was not loading on first load because the BEARER_TOKEN was `null` for some reason, we found that changing the `BEARER_TOKEN` constant to be a function fixed the issue, probably bc when the `BEARER_TOKEN` first gets its value upon page load, it doesn't update unless we refresh the page, but we want to get the access token value at the given moment

## Steps to test
1. Login
2. Go to access management page for the first time upon logging in
3. Data in the table should load without refreshing

* 

## Checklist

### DIFF
--- a/frontend/src/constants/AuthConstants.ts
+++ b/frontend/src/constants/AuthConstants.ts
@@ -2,12 +2,11 @@ import { getLocalStorageObjProperty } from "../utils/LocalStorageUtils";
 
 const AUTHENTICATED_USER_KEY = `${window.location.hostname}:AUTHENTICATED_USER`;
 
-export const getBearerToken = () => 
-{ 
+export const getBearerToken = (): string => {
   return `Bearer ${getLocalStorageObjProperty(
-  AUTHENTICATED_USER_KEY,
-  "accessToken",
-)}`; 
-}
+    AUTHENTICATED_USER_KEY,
+    "accessToken",
+  )}`;
+};
 
 export default AUTHENTICATED_USER_KEY;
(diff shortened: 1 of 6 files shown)


AmandaMisjuwar:
Also, again I'm not able to run the linter so if somebody who's reviewing this can run `yarn fix` on the frontend that would be great :DD

dhmi19 (approved):
lgtm. I'll let @e-wai take a look too since this impacts all our frontend calls

frontend/src/constants/AuthConstants.ts
```diff
@@ -2,9 +2,11 @@ import { getLocalStorageObjProperty } from "../utils/LocalStorageUtils";
 
 const AUTHENTICATED_USER_KEY = `${window.location.hostname}:AUTHENTICATED_USER`;
 
-export const BEARER_TOKEN = `Bearer ${getLocalStorageObjProperty(
-  AUTHENTICATED_USER_KEY,
-  "accessToken",
-)}`;
+export const getBearerToken = () => {
```
e-wai, line 5: linter

e-wai (approved):
good catch, can we fix the return type of function linter error before merging?

e-wai:
FWIW it's still very slow for me, will check on staging once this is merged

https://user-images.githubusercontent.com/40974372/204486567-1ad07fe8-a993-4b3e-83a9-c1b1c9d41cae.mov

PR #171, opened 2022-12-01
Implement camp creation footer component

twilkhoo:
## Notion ticket link

[Camp Creation: View only footer, back + next buttons](https://www.notion.so/uwblueprintexecs/Camp-Creation-View-only-footer-back-next-buttons-3d31f2dc04d042b8aa99bc4762189902)

## Implementation description
* Added a sticky footer, similar to what was done for the registration form.

https://user-images.githubusercontent.com/30396273/205034940-8f55314d-ddb1-494f-bcbf-9705d0aac731.mp4

## Steps to test
1. Go to the add camp page http://localhost:3000/admin/add-camp and test out the footer on the different pages.

* Correctness, logic.

## Checklist


frontend/src/components/pages/CampCreation/CampCreationFooter.tsx
```diff
@@ -0,0 +1,79 @@
...
+        disabled={currentStep === CampCreationPages.CampCreationDetailsPage}
+      >
+        Back
+      </Button>
+      <Button
+        width={{ sm: "95vw", lg: "auto" }}
+        height="48px"
+        variant="primary"
+        onClick={onNextStep}
+        disabled={!isCurrentStepCompleted}
```
twilkhoo, line 69: I noticed that for the registration form pr, we decided to have alerts when the next button was clicked but the page was incomplete. Right now I have it disabled as my ticket says, but should I turn these into alerts?

dhmi19: I think that was intentional. The interactions for camp creation and registration are slightly different. I think you can leave yours as is

frontend/src/components/pages/CampCreation/CampCreationFooter.tsx
```diff
@@ -0,0 +1,79 @@
...
+  handleStepNavigation: (stepsToMove: number) => void;
+};
+
+const CampCreationFooter = ({
+  currentStep,
+  isCurrentStepCompleted,
+  handleStepNavigation,
+}: CampCreationFooterProps): React.ReactElement => {
+  const onNextStep = () => {
+    if (isCurrentStepCompleted) {
```
dhmi19, line 17: I am not sure if I understand the logic here completely. When you're on the registrationFormPage step (the last step) and you click next, wouldn't you go to the next step (which doesn't exist)?

dhmi19: I guess the function is created in a way that it would do nothing if we increment past the last step? Also, wouldn't isCurrentStepCompleted always be true since the button would be disabled otherwise?

frontend/src/components/pages/CampCreation/CampCreationFooter.tsx
```diff
@@ -0,0 +1,79 @@
...
+  };
+
+  return (
+    <Flex
+      bg="white"
+      boxShadow="sm"
+      minH="92px"
+      width="100vw"
+      align="center"
+      justify={{ sm: "center", lg: "end" }}
```
dhmi19, line 33: Did you happen to get this from Bowen's PR? If so, we don't need to make the staff tool responsive so I think we can probs remove this?

dhmi19 (commented):
Functionality looks good but just had a few questions about the code

frontend/src/components/pages/CampCreation/index.tsx
```diff
@@ -19,20 +20,30 @@ const CampCreationPage = (): React.ReactElement => {
...
+  const [scheduledSessions, setScheduledSessions] = React.useState<CreateCampSession[]>([]);
 
   const [visitedRegistrationPage, setVisitedRegistrationPage] = useState(false);
 
   // Variables to determine whether or not all required fields have been filled out.
   // NOTE: This will depend on what type of state a page requires, i.e. determining
   // if a checkbox is checked is different than determining if an input field is filled.
-  const isCampDetailsFilled =
-    campDetailsDummyOne && campDetailsDummyTwo && campDetailsDummyThree !== "";
+  const isCampDetailsFilled = campDetailsDummyOne && campDetailsDummyTwo && campDetailsDummyThree !== "";
```
e-wai, line 30: Did linter not want two lines?

twilkhoo: I put that there before (just for my own sake) but now I've re-enabled the linter on those lines so it shouldn't be a problem anymore.

frontend/src/components/pages/CampCreation/CampCreationFooter.tsx
twilkhoo: Done.

frontend/src/components/pages/CampCreation/CampCreationFooter.tsx
twilkhoo: Looks like `onNextStep()` will call the index's `handleStepNavigation()`, and you're right that doesn't do anything if we're trying to move to a page that doesn't exist i.e. moving forward after the last page. After we implement the publish functionality, we can pass in a method to call the publish function as a prop to the footer.

Regarding your `isCurrentStepCompleted` comment, yeah you're right the if statement here is unneccesary, I removed it.

dhmi19 (approved):
lgtm 🚀

PR #175, opened 2022-12-02
Camp creation form

IvanYu327:
## Notion ticket link
[Camp Creation/Editing: Camp Details Form Fields](https://www.notion.so/Task-Board-F22-9e03bf9f7c4343f09a8a5c16dc48a12a?p=37e1bbe6fdbe431e93b71db2c89f2b36&pm=c)

## Implementation description
 - Created camp creation details form
 - Form fields filled validation
 - Image upload
 - Stored all input in state variables as outlined in ticket
 - Added a dummy submit button that acts as a submit so that unfilled form errors are only shown on submission attempt (rather than showing them all the time since the fields are always empty at the start and it doesnt make sense to show all the red messages then)

## Steps to test
1. Fill in the form
2. Press the dummy submit on an empty form
3. Fill in the fields and check if the errors disappear (and reappear if u empty the field again)
4. Upload an image via click and drop or drop
5. Try uploading non image files and files >5MB

Major Changes:
- state variables in the `CampCreation/index.tsx`
- `CampDetails/index.tsx`

## Checklist


frontend/src/components/pages/CampCreation/CampDetails/index.tsx
```diff
@@ -1,49 +1,520 @@
...
+  handleEarliestDropOffTime,
+  handleLatestPickUpTime,
+  handlePriceEDLP,  
+  handleAddressLine1,
+  handleAddressLine2,
+  handleCity,
+  handleProvince,
+  handlePostalCode,
+  setCampImageURL,
+}: CampCreationDetailsProps): JSX.Element => {
```
dhmi19, line 136: Could we use React.ReactElement here instead? I think JSX.Element is usually used for library components/ is more general purpose if I am not wrong. It handles types a bit differently too IIRC

dhmi19:
Little Bug:

https://user-images.githubusercontent.com/57159970/205415204-eeae4a8c-83a1-4a10-87c5-a034ed799dba.mov

Once you delete the contents of the field, the 0 doesn't go away

frontend/src/components/pages/CampCreation/CampDetails/index.tsx
```diff
@@ -1,49 +1,520 @@
...
+
+  const handleOnDrop = (event: React.DragEvent<HTMLDivElement>) => {
+    event.preventDefault(); 
+    event.stopPropagation(); 
+    const imageFile = event.dataTransfer.files[0];
+    handleFile(imageFile);
+  }
+
+
+  function getSectionTitle(number: number) {
```
dhmi19, line 188: I think this function is not being used, could we remove?

frontend/src/components/pages/CampCreation/CampDetails/index.tsx
```diff
@@ -1,49 +1,520 @@
...
+    const code = "A".charCodeAt(0);
+    return `Section ${String.fromCharCode(code + number)}`;
+  }
+
+
+
+  const {
```
dhmi19, line 195: Do we need these states?

dhmi19:
I think you have a nested paragraph component in there somewhere?
image

frontend/src/components/pages/CampCreation/index.tsx
```diff
@@ -11,13 +10,29 @@ import {
...
 } from "../../../types/CampsTypes";
+import CampCreationDetails from "./CampDetails";
 
 const CampCreationPage = (): React.ReactElement => {
   /* eslint-disable */
   // All response state from the three page components.
-  const [campDetailsDummyOne, setCampDetailsDummyOne] = useState(false);
-  const [campDetailsDummyTwo, setCampDetailsDummyTwo] = useState(false);
-  const [campDetailsDummyThree, setCampDetailsDummyThree] = useState("");
+  const [campName, setCampName] = useState<string>();
```
dhmi19, line 18: When the input changes, we currently get the following error:
image

Could you check if all the setters and states are correct? You might also want to use default values instead of undefined here (eg: empty strings, 0 for ints etc.)

frontend/src/components/pages/CampCreation/CampDetails/index.tsx
```diff
@@ -1,49 +1,520 @@
...
+      </HStack>
+      
+      <Checkbox marginTop="8px" onChange={toggleEDLP}>
+        <Text textStyle="buttonSemiBold">
+          Camp offers early drop-off and late pick-up
+        </Text>
+      </Checkbox>
+      
+
+      {offersEDLP?(
```
dhmi19, line 329: I think you can use `offersEDLP && ()` so you don't have the render the empty component with the ternary operator

frontend/src/components/pages/CampCreation/CampDetails/index.tsx
```diff
@@ -1,49 +1,520 @@
...
+                value={priceEDLP}
+                borderColor={!priceEDLP && showErrors?"red":"gray.200"}
+                borderWidth={!priceEDLP && showErrors?"2px":"1px"}
+                onChange={handlePriceEDLP}/>
+          </InputGroup>
+          {errorText(priceEDLP,"You must add a fee.")}
+        </>
+      ):(<></>)
+      }
+      <Box width="50%" maxWidth="870px" bg="#8C9196" height="2px" margin="40px 0"/>
```
dhmi19, line 379: would it be better to wrap the contents above in a box and then add margin to that? this seems a little fragile lol

frontend/src/components/pages/CampCreation/index.tsx
```diff
@@ -164,7 +262,8 @@ const CampCreationPage = (): React.ReactElement => {
         isScheduleSessionsFilled={isScheduleSessionsFilled}
         isRegistrationFormFilled={isRegistrationFormFilled}
       />
-      <Box w="100%" flex="1" mt={[0, "0 !important"]}>
+
+      <Box w="100%" flex="1" mt={[0, "0 !important"]} paddingLeft="200px">
```
dhmi19, line 266: Curious as to why you added more padding to the left here. Would this impact all the other steps too since it's on the parent component?

dhmi19 (changes requested):
This is looking really good and the image component is 🧑‍🍳 . Could you also run the linter with the following command:

`docker exec -it scv2_frontend /bin/bash -c "yarn fix"`

IvanYu327:
Fixed all the changes in the PR. However, since numeric state values (fees, age, etc) are initialized to 0, all the number fields start autopopulated with 0 since we're initial populating the input fields with the states (ie when a user opens a draft). I could implement logic/states to check if a 0 is a initial 0 or an entered 0 (like if the camp fee is 0 or the lower age is 0 which is technically possible) and then display it if needed, but not sure if that is an effort thats worth it right now.

dhmi19 (approved):
lgtm 🚢

PR #177, opened 2022-12-03
Create workflows for production deploys

e-wai:
## Implementation description

* Created new secrets for production apps in a production environment. Secrets are here: https://github.com/uwblueprint/focus-on-nature/settings/secrets/actions
  - [x] TODO: There's one secret missing, which is the service account JSON for the frontend deploy in production. I wasn't sure if that's something we wanted to store in raw JSON form -- I suspect we do right now, but hard to check. Firebase suggests doing it [docs](https://github.com/FirebaseExtended/action-hosting-deploy/blob/main/docs/service-account.md) but Github seems to warn against it [docs](https://docs.github.com/en/actions/security-guides/encrypted-secrets)
    > To help ensure that GitHub redacts your secret in logs, avoid using structured data as the values of secrets. For example, avoid creating secrets that contain JSON or encoded Git blobs.
  - [x] TODO: We should have a separate MongoDB instance for prod data. Not sure if we want this in the same project, since everyone currently has access?
  * Docs on environments: https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment

* Regenerated all staging backend credentials, and created new ones for production apps. All activity should be visible here (as are the links to both backend deploys): https://dashboard.heroku.com/apps
  * Uses my personal blueprint email and linked to my card, but we can swap this out later. It looks like we can transfer ownership on Heroku [docs](https://devcenter.heroku.com/articles/transferring-apps)

* Created a new `focus-on-nature-prod` project with a project alias so that the staging and production Firebase apps both live in this repo. This should make it clearer to people navigating our Firebase console (uses fon email, not personal)
  * https://firebase.google.com/docs/cli?authuser=1#use_aliases
  * Firebase recommends different environments point to different projects, I only think we need staging and production (https://firebase.google.com/docs/projects/dev-workflows/overview-environments, https://firebase.google.com/docs/projects/dev-workflows/general-best-practices)
  * Auth and mail service setup according to [starter code instructions](https://uwblueprint.github.io/starter-code-v2/docs/user-auth/)

* Other docs for context:
  * https://firebase.google.com/docs/hosting/github-integration
    - [ ] TODO: add preview links as authorized origins so that people can use OAuth (maybe pattern matching?) 
  * https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions
  * https://docs.github.com/en/actions/creating-actions/creating-a-composite-action#creating-an-action-metadata-file
  * Other bp examples:
    * https://github.com/uwblueprint/rowan-house/tree/main/.github
    * https://github.com/uwblueprint/ccbc/tree/development/.github
    - [ ] TODO: Figure out if we need a force push instead (eg. https://github.com/uwblueprint/rowan-house/commit/9311720c759a79ac87b8f02dc23c5a6091f9d757)
  * https://firebase.google.com/docs/reference/rest/auth/#section-sign-in-with-oauth-credential

## Steps to test
1. Not sure, but it's worth checking out the new project on https://console.firebase.google.com/ and https://console.cloud.google.com/
2. Check out Heroku dashboard

* Does the directory navigation make sense?
* Theoretically worth it to create individuals environments for staging and prod, I only made one for prod right now as to not break staging again. It does look like environment secrets take priority, so shouldn't ever conflict but I also didn't generate the initial credentials so it would just be a lot of unnecessary work
* Please read the TODOs, those are the spots I'm least sure about

## Checklist


.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,44 @@
...
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  build_and_deploy:
+    runs-on: ubuntu-latest
+    environment: fon-staging
```
e-wai, line 11: check whether `environment` is meaningful to us?

.github/workflows/deploy-prod.yml
```diff
@@ -0,0 +1,44 @@
...
+          firebase-storage-default-bucket: "${{ secrets.FIREBASE_STORAGE_DEFAULT_BUCKET }}"
+          firebase-project-id: "${{ secrets.FIREBASE_PROJECT_ID }}"
+          firebase-svc-account-private-key: "${{ secrets.FIREBASE_SVC_ACCOUNT_PRIVATE_KEY }}"
+          firebase-svc-account-client-email: "${{ secrets.FIREBASE_SVC_ACCOUNT_CLIENT_EMAIL }}"
+          mailer-user: "${{ secrets.MAILER_USER }}"
+          mailer-client-id: "${{ secrets.MAILER_CLIENT_ID }}"
+          mailer-client-secret: "${{ secrets.MAILER_CLIENT_SECRET }}"
+          mailer-refresh-token: "${{ secrets.MAILER_REFRESH_TOKEN }}"
+          stripe-publishable-key: "${{ secrets.STRIPE_PUBLISHABLE_TEST_KEY }}"
+          stripe-secret-key: "${{ secrets.STRIPE_SECRET_TEST_KEY }}"
```
e-wai, line 44: will need to identify which secrets are relevant to us

e-wai (commented):
Not sure how to test this one, thoughts @dhmi19?

PR #173, opened 2022-12-02
refactored createCamps and createCampSessions function

dhmi19:
## Notion ticket link

[Camp Creation: Refactor create camp functionality to include campSessions + FormQuestions](https://www.notion.so/uwblueprintexecs/Task-Board-F22-9e03bf9f7c4343f09a8a5c16dc48a12a?p=d80f8ccd4a1d486584d6c565b8de7204&pm=c)

## Implementation description
* used transactions in campCreation functionality to cleanup the rollbacks
* called the existing functions for session and form creation to make a 1 stop shop for camp creation (useful to avoid multiple calls form the FE)

## Steps to test
1. Use postman to test camp creation. You can use the body I posted below as an example. Make sure to use "form-data" and past this as the value of a key (where the key is called "data") 

## What should reviewers focus on?
* Is any other parts of the code indirectly affected by these changes? 
* I think we can remove any route which tries to create, updated, delete camp sessions and formQuestions since I think the only functionality that allows that is editing the camp? 

## Checklist
- [x] My PR name is descriptive and in imperative tense
- [x] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [x] I have run the appropriate linter(s)
- [ ] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR

PR #181, opened 2022-12-03
Refactor: Camp Create/Update should be single endpoint to handle all needs of frontend

dhmi19:
## Notion ticket link

[Camp Editing: Refactor UpdateCampById](https://www.notion.so/uwblueprintexecs/Task-Board-F22-9e03bf9f7c4343f09a8a5c16dc48a12a?p=81b6349211244333812a7c228c2d5a3f&pm=c)
[Camp Creation: Refactor create camp functionality to include campSessions + FormQuestions](https://www.notion.so/uwblueprintexecs/Task-Board-F22-9e03bf9f7c4343f09a8a5c16dc48a12a?p=d80f8ccd4a1d486584d6c565b8de7204&pm=c)

## Implementation description
* Added sessions to the main methods (`updateCampById` and `createCamp`) then passed the session down to any helper methods (optionally) 
* Update Camp deletes all the current camp sessions and then recreates them -> not an ideal implementation but it was easier/safer to refactor this way than messing with the `updateCampSessionsById` function.

## Steps to test
1. You can find steps to test camp creation in this closed [PR](https://github.com/uwblueprint/focus-on-nature/pull/173)
2. For update, first create a camp, and then use the body in the comment below make a PATCH request to `localhost:5000/camp/:campId`  

* Is the stripe stuff being create/updated correctly? Test going from inactive camp to active camp 
* When creating a draft camp, the user would not have filled out all the info, so we would need to send default values to the create function 

## Checklist


dhmi19:
Note: Refer to my comments about camp creation (including some tests) in the previous PR

### Body for updating camp
```
{
	"active": true,
	"ageLower": 9,
	"ageUpper": 12,
	"campCounsellors": [
		"63385e64f5716eea9ee462c6"
	],
	"campCoordinators": [
		"63385e64f5716eea9ee462c6",
		"634f046745baf02feaedd77a"
	],
	"capacity": 15,
	"description": "Focus on Nature camp hosted in Kitchener",
	"dropoffFee": 15,
	"startTime": "9:00",
	"endTime": "15:00",
	"earlyDropoff": "8:00",
	"fee": 15,
	"latePickup": "16:00",
	"location": {
		"streetAddress1": "456 Kitchen Blvd",
		"streetAddress2": "Unit 23",
		"city": "Kitchener",
		"province": "Ontario",
		"postalCode": "M4L 1M2"
	},
	"name": "Kitchener Camp B Dhruv Test Transaction Abort",
	"pickupFee": 20,
	"campSessions": [{
			"capacity": 20,
			"dates": [
				"Mon Feb 28 2022 08:00:00 GMT-0400 (Eastern Daylight Time)",
				"Tue Mar 1 2022 08:00:00 GMT-0400 (Eastern Daylight Time)",
				"Wed Mar 2 2022 08:00:00 GMT-0400 (Eastern Daylight Time)"
			]
		},
		{
			"capacity": 20,
			"dates": [
				"Mon Mar 7 2022 08:00:00 GMT-0400 (Eastern Daylight Time)",
				"Tue Mar 8 2022 08:00:00 GMT-0400 (Eastern Daylight Time)",
				"Wed Mar 9 2022 08:00:00 GMT-0400 (Eastern Daylight Time)"
			]
		},
		{
			"capacity": 20,
			"dates": [
				"Mon Mar 14 2022 08:00:00 GMT-0400 (Eastern Daylight Time)",
				"Tue Mar 15 2022 08:00:00 GMT-0400 (Eastern Daylight Time)",
				"Wed Mar 16 2022 08:00:00 GMT-0400 (Eastern Daylight Time)"
			]
		}
	],
	"formQuestions": [{
			"question": "this is question 1",
			"required": true,
			"category": "CampSpecific",
			"type": "Text"
		},
		{
			"question": "this is question 2",
			"required": true,
			"category": "CampSpecific",
			"type": "Text"
		}
	]
}
```

backend/typescript/services/implementations/campService.ts
```diff
@@ -1481,6 +1541,25 @@ class CampService implements ICampService {
     }
   }
 
+  async deleteFormQuestionsByIds(
+    formQuestionIds: string[],
+    dbSession?: mongoose.ClientSession,
```
e-wai, line 1546: sidenote, are helper methods supposed to be declared in the interface?

backend/typescript/services/implementations/campService.ts
```diff
@@ -551,7 +599,7 @@ class CampService implements ICampService {
       campCounsellors: newCamp.campCounsellors?.map((counsellor) =>
         counsellor.toString(),
       ),
-      campSessions: newCamp.campSessions?.map((session) => session.toString()),
+      campSessions: newCampSessions.map((cs) => cs.id),
```
e-wai, line 602: Do we need to cast to string? Is it not an `ObjectID`?

backend/typescript/services/interfaces/campService.ts
```diff
@@ -112,6 +117,17 @@ interface ICampService {
...
+  /**
+   * deleteFormQuestionsByIds deletes the formQuestions provided
+   * @param formQuestionIds the ids of the formQuestions to delete
+   * @param dbSession an optional session object used by parent functions if this is part of a series of changes
+   * @throws Error if formQuestions cannot be deleted
+   */
+  deleteFormQuestionsByIds(
+    formQuestionIds: string[],
+    dbSession?: mongoose.ClientSession,
+  ): Promise<void>;
```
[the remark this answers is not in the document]
dhmi19: yeah this is supposed to be a helper function only. You can never delete a formQuestion directly I think. The only use cases is to remove a question from a camp or remove it from the template. I'll make the dbSesssion mandatory

backend/typescript/types.ts
```diff
@@ -172,6 +172,8 @@ export type CreateCampDTO = Omit<
 > & {
   filePath?: string;
   fileContentType?: string;
+  formQuestions: CreateFormQuestionDTO[];
```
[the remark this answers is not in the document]
dhmi19: good catch, I think it's the same thing. I am gonna delete `CreateFormQuestionsDTO` since it's just an array of the base type

backend/typescript/services/implementations/campService.ts
dhmi19: hmm I am actually not sure. Should I remove them?

backend/typescript/services/implementations/campService.ts
dhmi19: newCampSessions is of type `CampSessionDTO[]`, so the camp session's ID is stored as a string there. Previously, we were using the model before but now I am using the result of a helper function

backend/typescript/services/implementations/campService.ts
e-wai: No idea, I'm happy to just keep it

e-wai (approved):
> When creating a draft camp, the user would not have filled out all the info, so we would need to send default values to the create function

So how do we differentiate between values that the user passed vs. default values we've inserted? (I think it's less than ideal but acceptable given how this code is shaped)

I still think that we need clearer distinction between what's intended to be called directly from routes, and what is only intended as a helper method. That's more of a high-level but I don't think it's worth blocking on since the functionality is all there

PR #187, opened 2022-12-07
Feat: Populates initial data for Camp Edit

dhmi19:
## Notion ticket link

[Camp Creation/Editing: Entry to editing flow](https://www.notion.so/uwblueprintexecs/Task-Board-F22-9e03bf9f7c4343f09a8a5c16dc48a12a?p=7c29108f44b442d79df40106b198b097&pm=c)
[Camp Edit: Populate initial camp data for edit flow](https://www.notion.so/uwblueprintexecs/Task-Board-F22-9e03bf9f7c4343f09a8a5c16dc48a12a?p=79665cc09fe148ca9c8a20fa795ab8cf&pm=c)

## Implementation description
* nothing crazy, just made a backend call to get the camps and then set the state of the camp creation flow to that data 

## Steps to test
1. First check the 2 entry points to edit camp: through the camp list table + through camp overview footer
2. Use the Vancouver camp as a test camp or create another draft camp to play around with
3. All the info should be preloaded on the edit-camp endpoint

* The comment on changing the key to change the defaultValue describes a (decently problematic) issue. I couldn't think of another way to handle the defaultValue changing 

## Checklist

### DIFF
--- a/frontend/src/components/pages/CampCreation/CampDetails/index.tsx
+++ b/frontend/src/components/pages/CampCreation/CampDetails/index.tsx
@@ -13,6 +13,7 @@ import {
   Image,
 } from "@chakra-ui/react";
 import IconImage from "../../../../assets/icon_image.svg";
+import { MAX_CAMP_DESC_LENGTH } from "../../../../constants/CampManagementConstants";
 
 type CampCreationDetailsProps = {
   campName: string;
@@ -174,7 +175,7 @@ const CampCreationDetails = ({
       {errorText(campName, "You must add a name.")}
 
       <Text textStyle="buttonSemiBold" marginTop="24px">
-        Short Description (max 400 characters){" "}
+        Short Description (max ${MAX_CAMP_DESC_LENGTH.toString()} characters){" "}
         <Text as="span" textStyle="buttonSemiBold" color="red">
           *
         </Text>
@@ -182,7 +183,7 @@ const CampCreationDetails = ({
       <Textarea
         width="575px"
         marginTop="8px"
-        maxLength={400}
+        maxLength={MAX_CAMP_DESC_LENGTH}
         defaultValue={campDescription}
         borderColor={!campDescription && showErrors ? "red" : "gray.200"}
         borderWidth={!campDescription && showErrors ? "2px" : "1px"}
@@ -204,7 +205,6 @@ const CampCreationDetails = ({
           type="number"
           placeholder="0.00"
           defaultValue={dailyCampFee}
-          key={dailyCampFee ? "loaded-camp-fee" : "loading-camp-fee"}
           borderColor={!dailyCampFee && showErrors ? "red" : "gray.200"}
           borderWidth={!dailyCampFee && showErrors ? "2px" : "1px"}
           onChange={handleDailyCampFee}
@@ -281,7 +281,6 @@ const CampCreationDetails = ({
                 height="52px"
                 maxLength={2}
                 defaultValue={ageLower}
-                key={ageLower ? "loaded-age-lower" : "loading-age-lower"}
                 borderColor={!ageLower && showErrors ? "red" : "gray.200"}
                 borderWidth={!ageLower && showErrors ? "2px" : "1px"}
                 onChange={handleAgeLower}
@@ -296,7 +295,6 @@ const CampCreationDetails = ({
                 height="52px"
                 maxLength={2}
                 defaultValue={ageUpper}
-                key={ageUpper ? "loaded-age-upper" : "loading-age-upper"}
                 borderColor={!ageUpper && showErrors ? "red" : "gray.200"}
                 borderWidth={!ageUpper && showErrors ? "2px" : "1px"}
                 onChange={handleAgeUpper}
@@ -320,9 +318,6 @@ const CampCreationDetails = ({
               height="52px"
               marginTop="8px"
               defaultValue={campCapacity}
-              key={
-                campCapacity ? "loaded-camp-capacity" : "loading-camp-capacity"
-              }
               borderColor={!campCapacity && showErrors ? "red" : "gray.200"}
               borderWidth={!campCapacity && showErrors ? "2px" : "1px"}
               onChange={handleCampCapacity}
@@ -411,7 +406,6 @@ const CampCreationDetails = ({
               type="number"
               placeholder="0.00"
               defaultValue={priceEDLP}
-              key={priceEDLP ? "loaded-priceEDLP" : "loading-priceEDLP"}
               borderColor={!priceEDLP && showErrors ? "red" : "gray.200"}
               borderWidth={!priceEDLP && showErrors ? "2px" : "1px"}
               onChange={handlePriceEDLP}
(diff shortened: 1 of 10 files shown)


frontend/src/components/pages/CampCreation/CampDetails/index.tsx
```diff
@@ -280,6 +281,7 @@ const CampCreationDetails = ({
                 height="52px"
                 maxLength={2}
                 defaultValue={ageLower}
+                key={ageLower ? "loaded-age-lower" : "loading-age-lower"}
```
dhmi19, line 284: I don't like this solution, but essentially defaultValue is only set on the first render (where the default values like ageLower = 0 is used). So afterwards, when the getCampById call returns us the current progress of a draft camp, the defaultValue doesn't update even though the underlying state updates. To solve this, we can change the key of the component whenever the data finishes loading. One of the issues with this is that the component looses focus every time we go from a truthy to falsy state and vice versa.

reference: https://stackoverflow.com/questions/30146105/react-input-defaultvalue-doesnt-update-with-state

frontend/src/components/pages/CampCreation/CampDetails/index.tsx
```diff
@@ -174,15 +174,15 @@ const CampCreationDetails = ({
       {errorText(campName, "You must add a name.")}
 
       <Text textStyle="buttonSemiBold" marginTop="24px">
-        Short Description (max 200 characters){" "}
+        Short Description (max 400 characters){" "}
```
e-wai, line 177: super nit: let's make this constant somewhere?

e-wai (changes requested):
Screen Shot 2022-12-11 at 2 02 00 AM

Should the "Save as draft" button be enabled when editing? in `CampCreationFooter.tsx`:

https://github.com/uwblueprint/focus-on-nature/blob/6a4134da17a7b3e2e71d40c3f96953c6bb58f7d3/frontend/src/components/pages/CampCreation/CampCreationFooter.tsx#L39-L47

frontend/src/components/pages/CampCreation/CampDetails/index.tsx
e-wai: To be fair, this was endorsed by the React docs - https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key

I forced reload on a parent element, does this (https://github.com/uwblueprint/focus-on-nature/commit/786d56c0eacb2b7f9d3d2435a16657192f9618ed) look any better? Feel free to `git cherry-pick` that if it helps

frontend/src/components/pages/CampCreation/index.tsx
```diff
@@ -159,6 +165,89 @@ const CampCreationPage = (): React.ReactElement => {
...
+          setCampCapacity(editCamp.campSessions[0].capacity);
+        }
+      }
+    };
+
+    // We only want to fetch the current state of the camp if we are editing a draft camp
+    if (editCampId) {
+      setInitialEditState();
+    }
+  }, [editCampId]);
```
e-wai, line 249: I don't think we need the `editCampId`? 
- It should re-evaluate if the URL is changed IIRC and never change while on the same camp
- Can combine this with the other `useEffect()`

dhmi19: The linter was complaining about the dependancy being missing so I added it. I don't expect this to change whilst on the same url so thought it wouldn't make a difference. As for combining the use effects, I thought it was more readable to separate them. Otherwise, the singular useEffect would be really long and hard to understand. This way, we can isolate the logic for populating the initial edit data. thoguhts?

frontend/src/components/pages/CampCreation/index.tsx
```diff
@@ -239,7 +328,11 @@ const CampCreationPage = (): React.ReactElement => {
       case CampCreationPages.ScheduleSessionsPage:
         return (
           <ScheduleSessions
-            campTitle="Waterloo Photography Camp 2022 @7:00 AM - 3:00PM"
+            campTitle={
+              campToEdit
+                ? `${campToEdit.name} @${campToEdit.startTime} - ${campToEdit.endTime}`
+                : "Waterloo Photography Camp 2022 @7:00 AM - 3:00PM"
```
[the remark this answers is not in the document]
dhmi19: I think I can actually just use the campName and states that we defined. On camp edit, the states will be populated with the initial data anyways. Also, the campName, startTime, and endTime will always be non zero values because you can't go to the session scheduling step without filling out the first step

frontend/src/utils/CampUtils.ts
```diff
@@ -138,6 +138,48 @@ export const getFormattedCampDateRange = (
...
+      return "Sa";
+  }
+};
+
+// getSelectedWeekDaysFromDates takes an array of Dates and returns the weekdays that are part of this
+// session of camp. Days without camp for this session will have a value of false
+export const getSelectedWeekDaysFromDates = (
+  sessionDates: Date[],
+): Map<string, boolean> => {
+  const tempWeekDays = new Map<string, boolean>([
```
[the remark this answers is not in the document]
dhmi19: think I forgot to rename it at the end. changed it now

frontend/src/components/pages/CampCreation/CampDetails/index.tsx
dhmi19: ahhh yeah this works soo much better. thanks!

dhmi19:
> Screen Shot 2022-12-11 at 2 02 00 AM
> 
> Should the "Save as draft" button be enabled when editing? in `CampCreationFooter.tsx`:
> 
> https://github.com/uwblueprint/focus-on-nature/blob/6a4134da17a7b3e2e71d40c3f96953c6bb58f7d3/frontend/src/components/pages/CampCreation/CampCreationFooter.tsx#L39-L47
> 
> Other comments:
> 
> * Need better handling for invalid camp id (request fails), I'm getting an error
> * Need better state for accessing valid camp id but camp is not editable (eg. published)

I could redirect to camp list page, but honestly, I don't see how the user would go to that URL unless they manually type it in.

e-wai:
> I could redirect to camp list page, but honestly, I don't see how the user would go to that URL unless they manually type it in.

I guess they could very feasibly typo the URL, so I would still push for handling that case -- should be a pretty minor change anyways

e-wai (approved):
Code looks great! I'm leaving an approval if this is urgent, but I would prefer handling the error case. I also want to test some more manually -- can we enable the "Save as draft" button in the footer in camp creation? it makes it really difficult to test the flow otherwise

dhmi19:
> Code looks great! I'm leaving an approval if this is urgent, but I would prefer handling the error case. I also want to test some more manually -- can we enable the "Save as draft" button in the footer in camp creation? it makes it really difficult to test the flow otherwise

I'll fix the URL error handling. This PR is only to load the initial data into the camp edit flow so the "save as draft" button is not implemented in this PR

dhmi19:
Closing this PR as the new PR includes these changes and is the latest version of create/edit camps

PR #196, opened 2022-12-14
Feat: Ability to create and edit camps

dhmi19:
## Notion ticket link

[Camp Edit: Populate initial camp data for edit flow](https://www.notion.so/uwblueprintexecs/Task-Board-F22-9e03bf9f7c4343f09a8a5c16dc48a12a?p=79665cc09fe148ca9c8a20fa795ab8cf&pm=c)

[Camp Editing: Publish and Save as Draft for Draft camps](https://www.notion.so/uwblueprintexecs/Task-Board-F22-9e03bf9f7c4343f09a8a5c16dc48a12a?p=c91afbd9f843439097fe59906a87f9a7&pm=c)

[Camp Creation: Publish and Save as Draft new camps](https://www.notion.so/uwblueprintexecs/Task-Board-F22-9e03bf9f7c4343f09a8a5c16dc48a12a?p=9cb856bf47e2431192ab71cbd456d5f1&pm=c)

## Implementation description
* 

## Steps to test
1. Have fun with the UI. You should be able to do everything with camp creation and editing now

* Is the Stripe stuff setup correctly (I edited some code but I think the functionality should remain the same)

## Checklist


backend/typescript/services/implementations/campService.ts
```diff
@@ -440,37 +442,72 @@ class CampService implements ICampService {
...
-            latePickup: camp.latePickup,
-            dropoffFee: camp.dropoffFee,
-            pickupFee: camp.pickupFee,
-            location: camp.location,
-            startTime: camp.startTime,
-            endTime: camp.endTime,
-            volunteers: camp.volunteers,
-            fee: camp.fee,
-            fileName: camp.filePath || oldCamp.fileName ? fileName : null,
+      if (
```
bowenzhu1, line 445: do we need to separate this? can we just combine and assign earlyDropoff, etc. even if theyre undefined

backend/typescript/services/implementations/campService.ts
```diff
@@ -537,24 +574,26 @@ class CampService implements ICampService {
 
         const campFee = newCamp.fee;
 
-        const dropoffPriceObject = await createStripePrice(
-          dropoffProductId,
-          dropoffFee * 100,
-        );
+        if (pickupFee && dropoffFee) {
```
bowenzhu1, line 577: What about if only one if valid? It seems like this wouldn't handle just a single payment (ex. just pickupFee)

backend/typescript/services/implementations/campService.ts
```diff
@@ -1291,8 +1330,9 @@ class CampService implements ICampService {
       const stripeDropoffProduct = await createStripeDropoffProduct(camp.name);
       const stripePickupProduct = await createStripePickupProduct(camp.name);
 
+      // Create the dropoff
       let priceIds = { dropoffPriceId: "", pickupPriceId: "" };
-      if (camp.active) {
+      if (camp.active && camp.dropoffFee && camp.pickupFee) {
```
bowenzhu1, line 1335: Same here as above - is it guaranteed that both fees are paid together? I assumed a parent can opt in for just one

backend/typescript/services/implementations/campService.ts
```diff
@@ -537,24 +574,26 @@ class CampService implements ICampService {
 
         const campFee = newCamp.fee;
 
-        const dropoffPriceObject = await createStripePrice(
-          dropoffProductId,
-          dropoffFee * 100,
-        );
+        if (pickupFee && dropoffFee) {
+          const dropoffPriceObject = await createStripePrice(
```
bowenzhu1, line 578: are these objects necessary? looks like we only need the IDs here

backend/typescript/types.ts
```diff
@@ -169,29 +169,37 @@ export type CreateCampDTO = Omit<
...
-  | "dropoffProductId"
-  | "pickupPriceId"
-  | "pickupProductId"
-> & {
-  filePath?: string;
-  fileContentType?: string;
-  campSessions: CreateCampSessionsDTO;
-  formQuestions: CreateFormQuestionDTO[];
-};
+// export type UpdateCampDTO = Omit<
```
bowenzhu1, line 187: what are we keeping this for? maybe add a comment on top for context

frontend/src/components/pages/CampCreation/index.tsx
```diff
@@ -159,87 +170,254 @@ const CampCreationPage = (): React.ReactElement => {
...
+        setAgeLower(editCamp.ageLower);
+        setAgeUpper(editCamp.ageUpper);
+        setOffersEDLP(
+          editCamp.earlyDropoff.length !== 0 ||
+            editCamp.latePickup.length !== 0,
+        );
+        setEarliestDropOffTime(editCamp.earlyDropoff);
+        setLatestPickUpTime(editCamp.latePickup);
+        setPriceEDLP(editCamp.pickupFee);
+        setAddresLine1(editCamp.location.streetAddress1);
```
bowenzhu1, line 233: typo here address

backend/typescript/services/implementations/campService.ts
bowenzhu1: Actually I see that EDLP opt in is all in one, so perhaps disregard the above.

frontend/src/components/pages/CampCreation/index.tsx
```diff
@@ -159,87 +170,254 @@ const CampCreationPage = (): React.ReactElement => {
...
+        toast({
+          description: `An error occurred with ${
+            isNewCamp ? "creating" : "updating"
+          } ${campFields.name}. Please try again.`,
+          status: "error",
+          variant: "subtle",
+          duration: 3000,
+        });
+      }
+    } catch (error: unknown) {
```
bowenzhu1, line 318: Do we want the error toast to be more informative? I tried Save as draft and it failed, and the error is: "Error: You must fill out all the required fields in step 1 and schedule at least 1 session before creation a camp. Please try again." 

Feels like saving as draft should be allowed even without scheduling camp sessions though?

frontend/src/utils/CampUtils.ts
```diff
@@ -138,6 +141,49 @@ export const getFormattedCampDateRange = (
...
+    ["Tu", false],
+    ["We", false],
+    ["Th", false],
+    ["Fr", false],
+    ["Sa", false],
+  ]);
+
+  sessionDates.forEach((date) => {
+    const key = getWeekDayKeyFromDateNum(date.getDay());
+    campDays.set(key, true);
```
bowenzhu1, line 181: this should be value right

bowenzhu1 (changes requested):
this is a pretty large PR with some pretty large files (1500+ lines damn) - I'm def still lacking context/understanding even after going through it, so a lot of comments are more like questions.

I think overall it looks good. But there's some minor changes + some clarification needed on behaviour I put

backend/typescript/middlewares/validators/campValidators.ts
```diff
@@ -75,30 +75,50 @@ export const createCampDtoValidator = async (
   if (!validatePrimitive(body.description, "string")) {
     return res.status(400).send(getApiValidationError("description", "string"));
   }
-  if (!validatePrimitive(body.earlyDropoff, "string")) {
+  if (body.earlyDropoff && !validatePrimitive(body.earlyDropoff, "string")) {
```
e-wai, line 78: I'm pretty sure `validatePrimitive` handles the null/undefined checks?

backend/typescript/services/implementations/campService.ts
e-wai: +1 to this, what happens if they have early dropoff but not late pickup? Can we use a conditional like https://github.com/uwblueprint/focus-on-nature/blob/6a4134da17a7b3e2e71d40c3f96953c6bb58f7d3/backend/typescript/services/implementations/campService.ts#L832?

edit: I think it has to have both - if so, can we leave a quick comment here?

frontend/src/components/common/NavBar.tsx
```diff
@@ -115,6 +115,7 @@ const NavBar = (): JSX.Element => {
                   color: "primary.green.100",
                   fontWeight: FontWeights.SEMIBOLD,
                 }}
+                onClick={() => history.push(Routes.CAMPS_PAGE)}
```
e-wai, line 118: tysm for this

frontend/src/components/pages/CampCreation/CampCreationFooter.tsx
```diff
@@ -6,18 +6,41 @@ export type CampCreationFooterProps = {
...
     if (currentStep === CampCreationPages.RegistrationFormPage) {
-      // eslint-disable-next-line no-console
-      console.log("Publish camp."); // Temporary.
+      // Publishing a new camp
+      if (!isEditingCamp) {
+        createUpdateCamp(true, true);
+      } else {
+        // Publishing a draft camp (should update the camp, not create a new camp)
+        createUpdateCamp(true, false);
+      }
```
e-wai, line 33: I would push for these arguments to be a type, it's pretty difficult to parse what's these mean when reading the code below

Alternatively, an easier fix would be something like:
```ts
const isPublished = true;
const isNewCamp = !isEditingCamp;

createUpdateCamp(isPublished, isNewCamp);
```

frontend/src/components/pages/CampCreation/CampDetails/index.tsx
```diff
@@ -174,15 +175,15 @@ const CampCreationDetails = ({
       {errorText(campName, "You must add a name.")}
 
       <Text textStyle="buttonSemiBold" marginTop="24px">
-        Short Description (max 200 characters){" "}
+        Short Description (max ${MAX_CAMP_DESC_LENGTH.toString()} characters){" "}
```
e-wai, line 178: nit: is the toString needed?

frontend/src/components/pages/CampCreation/CampDetails/index.tsx
```diff
@@ -207,6 +208,7 @@ const CampCreationDetails = ({
           borderColor={!dailyCampFee && showErrors ? "red" : "gray.200"}
           borderWidth={!dailyCampFee && showErrors ? "2px" : "1px"}
           onChange={handleDailyCampFee}
+          onWheel={(event) => event.currentTarget.blur()}
```
e-wai, line 211: oh interesting very cool

frontend/src/components/pages/CampCreation/index.tsx
```diff
@@ -159,87 +170,254 @@ const CampCreationPage = (): React.ReactElement => {
...
+        );
+        setEarliestDropOffTime(editCamp.earlyDropoff);
+        setLatestPickUpTime(editCamp.latePickup);
+        setPriceEDLP(editCamp.pickupFee);
+        setAddresLine1(editCamp.location.streetAddress1);
+        setAddresLine2(editCamp.location.streetAddress2 ?? "");
+        setCity(editCamp.location.city);
+        setProvince(editCamp.location.province);
+        setPostalCode(editCamp.location.postalCode);
+        setCampImageURL(editCamp.campPhotoUrl);
```
e-wai, line 238: would we be safe to make this an actual camp object eg. `CampResponse`? (I'm hoping enough of the flow is filled in to the point that this would be a useful thing to do, and reduce some of the repetitiveness of the getters/setters)

frontend/src/utils/CampUtils.ts
```diff
@@ -159,3 +205,88 @@ export const getSessionBorderColor = (sessionIndex: number): string =>
...
+    camp.campSessions.length > 0 &&
+    camp.campSessions.every((cs) => cs.capacity > 0)
+  ) {
+    return true;
+  }
+  return false;
+};
+
+// createUpdateCamp creates or updates a camp with the given fields
+export const createUpdateCamp = async (
```
e-wai, line 256: I think this belongs in CampsAPIClient, but I also don't see the huge benefit of having this as a combined method, since so much logic is based on the `isNewCamp` boolean

frontend/src/utils/CampUtils.ts
```diff
@@ -159,3 +205,88 @@ export const getSessionBorderColor = (sessionIndex: number): string =>
...
+    camp.location.city &&
+    camp.location.province &&
+    camp.location.province &&
+    camp.location.postalCode &&
+    camp.campSessions.length > 0 &&
+    camp.campSessions.every((cs) => cs.capacity > 0)
+  ) {
+    return true;
+  }
+  return false;
```
e-wai, line 252: Can we just return the condition directly, without the explicit `if` and `return` statements?

frontend/src/components/pages/CampCreation/index.tsx
```diff
@@ -159,87 +170,254 @@ const CampCreationPage = (): React.ReactElement => {
...
+    };
+
+    try {
+      const campResponse = await createUpdateCamp(
+        campFields,
+        isNewCamp,
+        editCampId,
+      );
+
+      if (campResponse) {
```
e-wai, line 298: Because we do: 
```ts
return error as CreateUpdateCampResponse;
```

I would guess this evaluates to true no matter success/error, can we check for `campResponse.id` or type instead? (I may be completely wrong)

backend/typescript/services/implementations/campService.ts
```diff
@@ -440,37 +442,72 @@ class CampService implements ICampService {
...
+              fileName: camp.filePath || oldCamp.fileName ? fileName : null,
+            },
+          },
+          {
+            new: true,
+            session,
+          },
+        ).populate({
+          path: "campSessions",
+          model: MgCampSession,
```
e-wai, line 508: how does this handle updates to formQuestions?

e-wai (changes requested):
what a legendddddddddd 😍 

mostly initial round of comments, will do some more manual testing but ty, this is amazing

backend/typescript/services/implementations/campService.ts
dhmi19: there will always be both fees if either is present. This is some tech debt from before that I would love to clean up but essentially pickupFee is always the same as dropoffFee so idk why they are seperate fields. As such, if one is present, the other should be too

backend/typescript/services/implementations/campService.ts
dhmi19: I didn't want to refactor the stripe API code in this PR since it's already wayyyyyy too big. I think we can revisit these improvements in future PRs?

frontend/src/components/pages/CampCreation/index.tsx
dhmi19: that's a requirement we have because the camp list table uses the sessions to determine which year a camp belongs to. without any sessions, the camp would not show up in the camp list page on the staff tool. The clients already agreed to this requirement so we should be good

frontend/src/utils/CampUtils.ts
dhmi19: I am using this function: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/set

this essentially sets the day to be true if it has a workshop scheduled on that day (hence true being hardcoded)

backend/typescript/middlewares/validators/campValidators.ts
dhmi19: Added the existance check because earlyDropoff is not a required field. if we only use the validatePrimitive then it would always fail for camps that don't offer EDLP since it won't be a type string

frontend/src/components/pages/CampCreation/index.tsx
dhmi19: it would but i didn't want to spend time on it- don't think it's worth making the code beautiful at this point honestly

frontend/src/utils/CampUtils.ts
dhmi19: just wanted to reduce code duplication since the code is pretty much the same but the text and call is conditionally rendered

backend/typescript/middlewares/validators/campValidators.ts
```diff
@@ -75,30 +75,50 @@ export const createCampDtoValidator = async (
...
+  if (body.pickupFee && !validatePrimitive(body.pickupFee, "number")) {
     return res.status(400).send(getApiValidationError("pickupFee", "number"));
   }
-  if (!validatePrimitive(body.dropoffFee, "number")) {
+  if (body.dropoffFee && !validatePrimitive(body.dropoffFee, "number")) {
     return res.status(400).send(getApiValidationError("dropoffFee", "number"));
   }
+  if (
+    (body.earlyDropoff && !body.dropoffFee) ||
+    (body.dropoffFee && !body.earlyDropoff)
```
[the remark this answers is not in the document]
dhmi19: this is saying that if you provide a time then you must provide a fee. alternatively, if you provide a fee, then you must provide a time. I'll change the error message to better reflect this

backend/typescript/types.ts
dhmi19: removed

frontend/src/utils/CampUtils.ts
dhmi19: having just a return statement requires explicitly checking for empty string and 0 for nums etc. since the && operator doesn't work with different types. Thought at that point that the if statement was cleaner

frontend/src/components/pages/CampCreation/index.tsx
dhmi19: I just fixed the campsAPI functions for create and update to actually throw errors so we can catch them properly.

backend/typescript/services/implementations/campService.ts
dhmi19: We do this for updating formQuestions:

```
// Update the FormQuestions by deleting all the current questions and creating new ones
    const currFormQuestions = (newCamp.formQuestions as FormQuestion[]).map(
      (fq) => fq.toString(),
    );
    await this.deleteFormQuestionsByIds(currFormQuestions, session);
    newFormQuestions = await this.addFormQuestionsToCamp(
      newCamp.id,
      camp.formQuestions,
      session,
    );
```

Although I think we can remove this populate functionality since these camp sessions are no longer the updated camp sessions

frontend/src/APIClients/CampsAPIClient.ts
```diff
@@ -31,27 +34,47 @@ const getCampById = async (id: string): Promise<CampResponse> => {
...
+  camp: CreateUpdateCampRequest,
+): Promise<CreateUpdateCampResponse> => {
+  // Check if atleast the first step and 1 session is scheduled
+  if (!isMinCampDetailsFilled(camp)) {
+    throw new Error(
+      `You must fill out all the required fields in step 1 and schedule at least 1 session before updating a camp`,
+    );
   }
+
+  const fieldsToUpdate: CreateUpdateCampRequest = { ...camp };
```
e-wai, line 67: why is this necessary?

frontend/src/components/pages/CampCreation/index.tsx
e-wai: So the create/edit call may throw an error with an message -- should that be displayed here? Could also be an alert if that's easier

frontend/src/components/pages/CampsList/CampsTable.tsx
```diff
@@ -215,6 +225,16 @@ const CampsTable = ({
                         Delete camp
```
e-wai, line 225: I think delete belongs in the popover footer

frontend/src/components/pages/CampCreation/CampCreationFooter.tsx
```diff
@@ -6,19 +6,53 @@ export type CampCreationFooterProps = {
...
-      console.log("Publish camp."); // Temporary.
+      // Publishing a new camp
+      if (!isEditingCamp) {
+        isPublishedCamp = true;
+        isNewCamp = true;
+      } else {
+        // Publishing a draft camp (should update the camp, not create a new camp)
+        isPublishedCamp = true;
+        isNewCamp = false;
+      }
```
e-wai, line 37: Can we streamline this and in `onSaveAsDraft`? still hard to read

```
const isPublishedCamp = true;
const isNewCamp = !isEditingCamp;
```

e-wai (approved):
Still some comments, but the functionality looks good to me -- only the flipping on/off EDLP boolean question is really blocking

backend/typescript/services/implementations/campService.ts
```diff
@@ -1291,8 +1330,9 @@ class CampService implements ICampService {
       const stripeDropoffProduct = await createStripeDropoffProduct(camp.name);
       const stripePickupProduct = await createStripePickupProduct(camp.name);
 
+      // Create the dropoff
```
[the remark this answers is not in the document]
dhmi19: this would not be a use case as confirmed with Andy

backend/typescript/services/implementations/campService.ts
```diff
@@ -537,27 +574,29 @@ class CampService implements ICampService {
 
         const campFee = newCamp.fee;
 
-        const dropoffPriceObject = await createStripePrice(
-          dropoffProductId,
-          dropoffFee * 100,
-        );
+        if (pickupFee && dropoffFee) {
```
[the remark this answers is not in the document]
dhmi19: I didn't change the logic of the stripe api stuff as I didn't have a complete understanding of the processes. I think it's worth merging as is for now and then refactoring if the testing fails.

frontend/src/components/pages/CampCreation/index.tsx
```diff
@@ -159,87 +168,254 @@ const CampCreationPage = (): React.ReactElement => {
...
+              }
+              handleAgeUpper={(event: React.ChangeEvent<HTMLInputElement>) =>
+                setAgeUpper(Number(event.target.value))
+              }
+              handleCampCapacity={(
+                event: React.ChangeEvent<HTMLInputElement>,
+              ) => setCampCapacity(Number(event.target.value))}
+              toggleEDLP={(event: React.ChangeEvent<HTMLInputElement>) => {
+                // If currently offers EDLP, reset the state of the timings to default values
+                if (offersEDLP) {
```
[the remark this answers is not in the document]
dhmi19: when you toggle it off, then this if condition would be true and you would reset to default values of edlp (ie: no timings and 0 fee)

frontend/src/components/pages/CampsList/CampsTable.tsx
dhmi19: pretty sure delete was in the popover too? This was unchanged from before though so I am not 100% sure

PR #206, opened 2022-12-21
Refactor waiver registration with error handling

TW1234567:
## Notion ticket link

[Waiver Registration: Add error message alerts](https://www.notion.so/uwblueprintexecs/Waiver-Registration-Add-error-message-alerts-a2aefff603554b31bb8cb41c3702a89a)

## Implementation description
* Leverage the `nextBtnRef` and a `useEffect` callback to notify any invalid input messages.
* Made it so that the optional clauses are not set to null instead of "I do not agree" by default

## Steps to test
1. Go to link such as `http://localhost:3000/register/camp/630a63450eb59852f9d01b1e`
2. Go to 3rd step and try to go to step 4 without fully filling out the waiver

* Proper usage of `nextBtnRef` and making sure that the event listeners are being correctly removed on component dismount.

## Checklist


frontend/src/components/pages/RegistrantExperience/Waiver/index.tsx
```diff
@@ -51,16 +58,63 @@ const boldButtonTextStyles = {
...
+  }, [waiverInterface.optionalClauses.length]);
+
+  useEffect(() => {
+    let nextBtnRefValue: HTMLButtonElement; // Reference to the next step button
+    const updateFormErrorMsgs = () => {
+      if (!checkName(waiverInterface.name)) setIsNameInvalid(true);
+      if (!checkDate(waiverInterface.date)) setIsDateInvalid(true);
+      if (!checkRequiredClauses(waiverInterface.agreedRequiredClauses))
+        setIsRequriedClauseInvalid(true);
+      const newIsOptionalClausesInvalid: boolean[] = Array.from(
```
dhmi19, line 97: Could we rewrite this check for optional clauses? It's a little confusing as to what we are trying to accomplish here. From what i understand, we create an array of booleans where each boolean refers to if the optional clause is NOT AGREED. Then we set invalidity based on this array. I am not sure what the underlying intention for this is

frontend/src/components/pages/RegistrantExperience/Waiver/index.tsx
```diff
@@ -51,16 +58,63 @@ const boldButtonTextStyles = {
...
+    );
+  }, [waiverInterface.optionalClauses.length]);
+
+  useEffect(() => {
+    let nextBtnRefValue: HTMLButtonElement; // Reference to the next step button
+    const updateFormErrorMsgs = () => {
+      if (!checkName(waiverInterface.name)) setIsNameInvalid(true);
+      if (!checkDate(waiverInterface.date)) setIsDateInvalid(true);
+      if (!checkRequiredClauses(waiverInterface.agreedRequiredClauses))
+        setIsRequriedClauseInvalid(true);
```
dhmi19, line 96: can we add curly braces around this if check? It's a little confusing as to what's part of the if block etc.

dhmi19 (changes requested):
Leaving some comments; functionality looks good though

frontend/src/types/waiverTypes.tsx
```diff
@@ -41,7 +42,7 @@ export enum WaiverActions {
 }
 
 export interface OptionalClauseResponse extends WaiverClause {
-  agreed: boolean;
+  agreed: boolean | null;
```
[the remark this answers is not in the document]
TW1234567: Yep! Let me remove it. Instead, I'm adding a field `optionSelected` to optionalClauseResponse. The field is used to show whether the parent has selected `I agree` or `I disagree` to a clause

frontend/src/components/pages/RegistrantExperience/Waiver/index.tsx
TW1234567: Yep!

frontend/src/components/pages/RegistrantExperience/Waiver/index.tsx
TW1234567: The variable `isOptionalClausesInvalid` stored an array of boolean where each Boolean refers to if the optional clause **has either been agreed or disagreed to**. Since the optional clauses are by default unfilled. We use `isOptionalClausesInvalid` to raise an error when clause is left unfilled. I'm not fully sure how to make the check clearer. The idea is that we create a new array of booleans by checking whether each clause is unfilled.

frontend/src/components/pages/RegistrantExperience/Waiver/index.tsx
TW1234567: Ok it seems that lint doesn't like that style. I will put the brackets instead as you suggested.

frontend/src/types/waiverTypes.tsx
```diff
@@ -17,6 +17,7 @@ export interface LoadedWaiver extends WaiverReducerDispatchBase {
 }
 export interface ClickOptionalClause extends WaiverReducerDispatchBase {
   type: WaiverActions;
+  agreed: boolean;
```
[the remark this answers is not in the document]
TW1234567: I feel hesitant to move the types in waiverTypes.tsx into AdminTypes file since the type in waiverTypes.tsx are specific to the waiverReducer which is specific only to the waiver registration. And Yep, let me rename the file.

dhmi19 (approved):

PR #117, opened 2022-10-01
feat: stripe payments

rchang0501:
## Notion ticket link

[Payments - Stripe API](https://www.notion.so/uwblueprintexecs/Payments-Stripe-API-5758698f7e8d440da53b92b84082df3c)

## Implementation description
Referencing [payments implementation doc on notion](https://www.notion.so/uwblueprintexecs/Payments-Implementation-0aa3f204372a426ca780823579e23226)
* created `createCampersCheckoutSession` function to create a stripe checkout session in camper service
* updated campers/register endpoint to call `createCampersCheckoutSession` and redirect to stripe hosted checkout page

## Steps to test
Method 1: 
1. make a post request at `http://localhost:5000/campers/register`
2. you should see the stripe page as a response 

method 1 sample request body: 
```
[
    {
        "campSession": "63139cdec3d7b55b44a01541",
        "firstName": "test",
        "lastName": "test",
        "age": 12,
        "allergies": "",
        "earlyDropoff": ["2022-04-06T00:00:00.000+00:00"],
        "latePickup": ["2022-04-06T00:00:00.000+00:00"],
        "specialNeeds": "", 
        "contacts": [
            {
                "firstName": "Mommy",
                "lastName": "?",
                "email": "mommyyy@gmail.com",
                "phoneNumber": "1234",
                "relationshipToCamper": "mommy"
            },
            {
                "firstName": "Daddy",
                "lastName": "?",
                "email": "daddy@gmail.com",
                "phoneNumber": "1234",
                "relationshipToCamper": "daddy"
            }
        ],
        "registrationDate": "2022-04-06T00:00:00.000+00:00",
        "hasPaid": false,
        "formResponses": {"622cfedaaf70bf090031d064":"five"}, 
        "chargeId": "hi", 
        "charges": {
            "camp": 20,
            "earlyDropoff": 0,
            "latePickup": 0
        },
        "optionalClauses": [
            {
                "clause": "",
                "agreed": true 
            }
        ]
    }
]
```

Method 2 (jenky frontend - tried to get the redirect to work with our camperAPIClient but gave up for time): 
1. checkout `feat/stripe-payments-FOR-TESTING` 
2. go to global forms page 
4. open your browser console 
5. press on the button 
6. click on the link with the "Access to XMLHttpReques" error 
7. fill out the stripe checkout page (reference https://stripe.com/docs/checkout/quickstart#testing for different credit cards to use) 

can change your test data in `frontend/src/components/pages/GlobalFormsPage.tsx`

[...]

PR #208, opened 2022-12-21
Implement payments flow

e-wai:
## Notion ticket link

[Registrant Experience: Payment Failure/Confirmation](https://www.notion.so/uwblueprintexecs/Task-Board-F22-9e03bf9f7c4343f09a8a5c16dc48a12a?p=c4a45ddbdb1f433cbf01448ede701316&pm=c)
[Payments - Stripe API](https://www.notion.so/uwblueprintexecs/Task-Board-F22-9e03bf9f7c4343f09a8a5c16dc48a12a?p=5758698f7e8d440da53b92b84082df3c&pm=c)

## Implementation description
* Used session storage to cache session details and restore session when the user clicks "back"
  * `sessionStorage` is different from `localStorage`, expires: https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage 
* Creates a bunch of components for result handling in `RegistrationResult/`

Based on work in https://github.com/uwblueprint/focus-on-nature/pull/117:
* To get around CORS errors for redirect, let's return the created checkout link and then initiate the movement to the stripe page in the frontend
* Tried adding this to the CORS allowlist but it wasn't enough, I suspect it's bc we can't control the Stripe origin?
* Also adds default camp image

The full e2e flow looks like:
* User goes through registration flow, clicks register
* Request sent to backend to create campers in db, and create checkout session in Stripe
  * If success, return the checkout session URL
  * If failed, return to checkout and show the checkout error modal
* Save session data (see `CheckoutData`) to session storage, including checkout URL, and redirect to checkout
  * note: handling for pressing back on webpage isn't great, is it worth using `window.location.replace()` instead of `window.location.assign()`?
* Stripe checkout
  * If payment success, Stripe redirects to `/register/camp/:campId?result=success`. We retrieve session storage for this page to load summary; if this fails, we have a generic message that says check email
  * If payment fails, Stripe handles that in the webpage (no redirect)
  * If payment cancelled (pressed back arrow on the Stripe checkout page), Stripe redirects to `/register/camp/:campId?result=cancel`. We restore session details from session storage and show error modal.

## Steps to test
1. Go through all checkout/payment flows using a Stripe test card (https://stripe.com/docs/testing)
  a. cancel (press back button) 
  b. complete payment -> should end up at success page
  c. incorrect card details / payment decline (see Stripe docs)
2. Look for success page states: through Stripe payment, also access directly and look at error state
3. Check mobile states for all of above (I think the modal is too wide on phone?)
4. Mess around with session Storage (invalidate it manually, see if app handles it)

* Look at `frontend/` only please -- backend created in separate PR https://github.com/uwblueprint/focus-on-nature/pull/212, and will rebase later

## Checklist


frontend/src/utils/RegistrationUtils.ts
```diff
@@ -0,0 +1,58 @@
...
+import { CampResponse } from "../types/CampsTypes";
+import { CartItem } from "../types/RegistrationTypes";
+
+export const getCheckoutSessionStorageKey = (campId: string): string =>
+  `checkout-${campId}`;
+
+export const getFailedSessionStorageKey = (campId: string): string =>
+  `failedCheckout-${campId}`;
+
+export const mapToCampItems = (
```
e-wai, line 11: TODO @e-wai

frontend/src/components/pages/RegistrantExperience/RegistrationResult/RegistrationErrorModal.tsx
```diff
@@ -0,0 +1,60 @@
...
+}: RegistrationErrorModalProps): React.ReactElement => {
+  return (
+    <Modal isOpen={isOpen} onClose={onClose} isCentered>
+      <ModalOverlay />
+      <ModalContent>
+        <ModalHeader>
+          <Text>{title}</Text>
+        </ModalHeader>
+        <ModalBody>
+          <Text px="8px">{message}</Text>
```
e-wai, line 37: not sure why there was different padding by default on header vs. body, can someone else check what this looks like on your screen?

frontend/src/APIClients/CamperAPIClient.ts
```diff
@@ -109,11 +111,23 @@ const getCampersByChargeIdAndSessionId = async (
   }
 };
 
+const registerCampers = async (
+  campers: CreateCamperRequest[],
+): Promise<CheckoutUrl> => {
+  const { data } = await baseAPIClient.post(`/campers/register`, campers, {
```
dhmi19, line 117: There are some changes to this backend endpoint: 
1. We now pass in an array of sessionIds to support multi-session registration
2. The actual campers array that Thomas created has changed so that the camper data that we send to the backend doesn't include the charges, id, and campSession fields. Just wanted to verify if those changes work with this.

Do you need the charges field on the campers for the stripe processing? I thought it would be cleaner for the backend to add the charges field than the frontend trying to calculate it

backend/typescript/rest/camperRoutes.ts
```diff
@@ -24,11 +24,10 @@ const camperService: ICamperService = new CamperService();
 camperRouter.post("/register", createCampersDtoValidator, async (req, res) => {
   try {
     const campers = req.body as CreateCampersDTO;
-    const newCampers = await camperService.createCampers(
+    const checkoutSessionUrl = await camperService.createCampersCheckoutSession(
```
dhmi19, line 27: I know I am not supposed to be commenting on the backend for this PR, but if we use this endpoint to create the checkout session, then at what point do we create the db entities? Should the checkout session creation be it's own GET endpoint `campers/checkoutSession`

frontend/src/components/pages/RegistrantExperience/index.tsx
```diff
@@ -109,62 +166,169 @@ const RegistrantExperiencePage = (): React.ReactElement => {
...
+    }
+  };
+
   const handleStepNavigation = (stepsToMove: number) => {
     const desiredStep = currentStep + stepsToMove;
     if (RegistrantExperienceSteps[desiredStep]) {
       setCurrentStep(currentStep + stepsToMove);
     } else if (desiredStep < 0) {
       alert("PLACEHOLDER - go to initial registration");
+    } else if (camp) {
```
dhmi19, line 201: I don't think I understand this condition. the first condition takes us to a defined page in the flow. The second takes us to the initial session selection page (i think?). Shouldn't this be a separate function called `handleCheckoutClick` or something which is conditionally called by the next/checkout button?

frontend/src/components/pages/RegistrantExperience/index.tsx
```diff
@@ -109,62 +166,169 @@ const RegistrantExperiencePage = (): React.ReactElement => {
...
+      registerCampers(registeringCampers, camp);
     } else {
-      alert("PLACEHOLDER - go to payment");
+      alert("Unexpected camp state. Please restart registration.");
```
dhmi19, line 204: Not sure how we would reach this state. If it's possible, I think we should rather avoid it than show this message as it's kinda confusing how the error even happened

frontend/src/components/pages/RegistrantExperience/index.tsx
```diff
@@ -109,62 +166,169 @@ const RegistrantExperiencePage = (): React.ReactElement => {
...
+            setErrorModalMessage({
+              title: "Payment cancelled",
+              body:
+                "No payment was processed. If this was not intentional, please try again.",
+            });
+            errorModalOnOpen();
+          }
+
+          sessionWasRestored = true;
+        } catch (err: unknown) {
```
dhmi19, line 247: Which function throws an error here? Can we avoid this?

frontend/src/components/pages/RegistrantExperience/index.tsx
```diff
@@ -109,62 +166,169 @@ const RegistrantExperiencePage = (): React.ReactElement => {
...
-          {camp ? (
-            getCurrentRegistrantStepComponent(currentStep)
-          ) : (
-            <Text>Error: Camp not found. Please go back and try again.</Text>
+    <>
+      {registrationResult === SUCCESS_RESULT_CODE ? (
+        <RegistrationResultPage
+          camp={camp}
+          campers={registeringCampers}
+          items={camp ? mapToCampItems(camp, registeringCampers) : undefined}
```
dhmi19, line 283: can the camp ever be undefined here? If so, would it make more sense to check for that in the top condition so we don't render the RegistrationResultPage until we have the camp data?

frontend/src/components/pages/RegistrantExperience/ReviewRegistration/index.tsx
```diff
@@ -1,62 +1,26 @@
...
-  campName,
+  campers,
+  camp,
 }: ReviewRegistrationProps): React.ReactElement => {
   return (
     <Box>
-      <PaymentSummary campName={campName} items={items} />
+      <PaymentSummary
+        campName={camp.name}
+        items={mapToCampItems(camp, campers)}
```
dhmi19, line 22: is it worth it to just map the camp to cart items once and pass around an array of cart items instead? seems like we are making a lot of calls to this function which could just be replaced by the state

frontend/src/types/CamperTypes.ts
```diff
@@ -34,6 +34,8 @@ export type Camper = {
   ];
 };
 
+export type CreateCamperRequest = Omit<Camper, "id">;
```
dhmi19, line 37: Could we use the `RegistrantExperienceCamper` type in dev instead? The multi-session backend PR made some changes to the data being passed in so this type would be more up to date (I pushed the commits to Thomas's step 1 PR  to make the types more consistent with my backend PR)

frontend/src/types/CamperTypes.ts
```diff
@@ -77,3 +79,5 @@ export type WaitlistedCamper = {
   campSession: string;
   linkExpiry?: Date;
 };
+
+export type CheckoutUrl = { checkoutSessionUrl: string };
```
dhmi19, line 83: why does this need to be an object?

frontend/src/components/pages/RegistrantExperience/index.tsx
```diff
@@ -109,62 +166,169 @@ const RegistrantExperiencePage = (): React.ReactElement => {
...
+
+      AdminAPIClient.getWaiver().then((waiver) => {
+        if (waiver.clauses) {
+          waiverDispatch({
+            type: WaiverActions.LOADED_WAIVER,
+            waiver,
+          });
+        }
+
+        // TODO error handling
```
dhmi19, line 272: did you want to work on these TODOs before merge?

dhmi19 (changes requested):
Looks good! just left a couple of comments. We should discuss the backend entity creation with the stripe processing. Also, I think we should send the parent and admin emails after the entities are created on the db and the payment has been processed?

frontend/src/types/CamperTypes.ts
e-wai: I guess it doesn't really, more of nice to have to see mark the expected response structure. Maybe renaming to `CheckoutUrlResponse` makes more sense?

frontend/src/components/pages/RegistrantExperience/ReviewRegistration/index.tsx
e-wai: Hmm maybe, but it feels like we're storing the same information redundantly though?

frontend/src/components/pages/RegistrantExperience/index.tsx
e-wai: should be resolved by session selection, ie. never go the registration steps without a camp

frontend/src/components/pages/RegistrantExperience/index.tsx
```diff
@@ -42,49 +73,75 @@ const RegistrantExperiencePage = (): React.ReactElement => {
...
+
+  const goToCheckout = (
+    checkoutSessionUrl: string,
+    currentCamp: CampResponse,
+  ) => {
+    saveSessionToSessionStorage(campId, {
+      campers: registeringCampers,
+      camp,
+      items: mapToCampItems(currentCamp, registeringCampers),
+      waiver: waiverInterface.waiver as WaiverType,
```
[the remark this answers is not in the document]
e-wai: Hmm it looked fine to me, but it does rely on the potentially dangerous assumption that camp is loaded by the time they're on the waiver page?

frontend/src/components/pages/RegistrantExperience/index.tsx
```diff
@@ -42,49 +73,75 @@ const RegistrantExperiencePage = (): React.ReactElement => {
...
+    sessionStorage.setItem(CAMP_ID_SESSION_STORAGE_KEY, sessionCampId);
+    sessionStorage.setItem(
+      getCheckoutSessionStorageKey(sessionCampId),
+      JSON.stringify(checkoutData),
+    );
+  };
+
+  const goToCheckout = (
+    checkoutSessionUrl: string,
+    currentCamp: CampResponse,
```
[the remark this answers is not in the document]
e-wai: ehh it's a stylistic thing that I've never fully resolved, whether methods should stay isolated from state dependencies and pass through different functions, especially since they can instead exist out of component. I guess in this case it reduces boilerplate by using state vars?

frontend/src/components/pages/RegistrantExperience/index.tsx
```diff
@@ -42,49 +73,75 @@ const RegistrantExperiencePage = (): React.ReactElement => {
...
-        return isPersonalInfoFilled;
-      case RegistrantExperienceSteps.AdditionalInfoPage:
-        return isAdditionalInfoFilled;
-      case RegistrantExperienceSteps.WaiverPage:
-        return isWaiverFilled;
-      case RegistrantExperienceSteps.ReviewRegistrationPage:
-        return isReviewRegistrationFilled;
-      default:
-        return false;
+  const [registeringCampers, setRegisteringCampers] = useState<
```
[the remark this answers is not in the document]
e-wai: will look at this in the rebase

frontend/src/types/CamperTypes.ts
dhmi19: yeah I think that would be clearer

PR #224, opened 2022-12-28
Add image upload to camp creation

e-wai:
## Notion ticket link

[Camp Creation/Editing: Upload Image & Store in Firebase ](https://www.notion.so/uwblueprintexecs/Task-Board-F22-9e03bf9f7c4343f09a8a5c16dc48a12a?p=0b7a1c33a79f47d6ad075fbf93b4ad50&pm=c)

## Implementation description

* The backend infra is set up already
* https://uwblueprint.github.io/starter-code-v2/docs/file-storage/
  *  https://cloud.google.com/storage/pricing#cloud-storage-always-free
* https://developer.mozilla.org/en-US/docs/Web/API/FormData/Using_FormData_Objects#sending_files_using_a_formdata_object
  * https://stackoverflow.com/questions/11876175/how-to-get-a-file-or-blob-from-an-object-url

## Steps to test
1.

* Camp creation and camp editing flows; keep or change image

## Checklist

### DIFF
--- a/frontend/src/components/pages/CampCreation/CampCreationFooter.tsx
+++ b/frontend/src/components/pages/CampCreation/CampCreationFooter.tsx
@@ -11,6 +11,7 @@ export type CampCreationFooterProps = {
     isPublishedCamp: boolean,
     isNewCamp: boolean,
   ) => Promise<void>;
+  setShowCreationErrors: React.Dispatch<React.SetStateAction<boolean>>;
 };
 
 const CampCreationFooter = ({
@@ -19,6 +20,7 @@ const CampCreationFooter = ({
   handleStepNavigation,
   createUpdateCamp,
   isEditingCamp,
+  setShowCreationErrors,
 }: CampCreationFooterProps): React.ReactElement => {
   const [isAwaitingReq, setIsAwaitingReq] = React.useState(false);
 
@@ -30,6 +32,10 @@ const CampCreationFooter = ({
   };
 
   const onNextStep = async () => {
+    if (!isCurrentStepCompleted) {
+      setShowCreationErrors(true);
+      return;
+    }
     handleStepNavigation(1);
 
     if (currentStep === CampCreationPages.RegistrationFormPage) {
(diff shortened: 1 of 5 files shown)


e-wai:
When we remove the dummySubmit button, we never have explicit checks for error states - other than disabling the "next" button (which also doesn't show us anything about the logical states). I'll put this in a separate PR

frontend/src/components/pages/CampCreation/CampDetails/index.tsx
```diff
@@ -95,8 +98,8 @@ const CampCreationDetails = ({
   handleProvince,
   handlePostalCode,
   setCampImageURL,
+  showErrors,
```
twilkhoo, line 101: Now you take `showErrors` as a boolean prop instead of state, so that the footer can set it to true/false.

frontend/src/theme/colors.ts
```diff
@@ -5,6 +5,9 @@ const colors = {
       200: "#3A7035",
       600: "#146600",
     },
+    green_disabled: {
+      100: "#B5CFB3",
```
twilkhoo, line 9: This color is the exact same as if we were to disable a primary.green.100 button.

frontend/src/components/pages/CampCreation/CampCreationFooter.tsx
```diff
@@ -79,7 +85,12 @@ const CampCreationFooter = ({
         height="48px"
         variant="primary"
         onClick={onNextStep}
-        disabled={!isCurrentStepCompleted}
```
twilkhoo, line 82: Now, the next button is never really _disabled_, but now instead looks disabled (with the change in background color) when some fields haven't been filled out. This was done so we still have access to the `onClick` function even when something hasn't been filled.

twilkhoo (approved):
I've tested your changes, everything looks good on my end! I added some frontend changes regarding error validation (since the dummy submit was the only thing that handled it before, but now we've removed it) so feel free to check out my changes if you have time.

PR #233, opened 2022-12-30
[1/n] Create confirm payment endpoint

e-wai:
## Notion ticket link

On payment success, update db to indicate that camper `hasPaid=true`

## Implementation description
* 

## Steps to test
1.

* 

## Checklist

### DIFF
--- a/backend/typescript/rest/camperRoutes.ts
+++ b/backend/typescript/rest/camperRoutes.ts
@@ -116,6 +116,20 @@ camperRouter.get("/:chargeId/:sessionId", async (req, res) => {
   }
 });
 
+// ROLES: unprotected
+/* On successful payment, mark camper as paid */
+camperRouter.post("/confirm-payment/:chargeId", async (req, res) => {
+  const { chargeId } = req.params;
+  try {
+    const camper = await camperService.confirmCamperPayment(
+      (chargeId as unknown) as string,
+    );
+    res.status(200).json(camper);
+  } catch (error: unknown) {
+    res.status(500).json({ error: getErrorMessage(error) });
+  }
+});
+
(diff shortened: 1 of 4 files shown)


backend/typescript/rest/camperRoutes.ts
```diff
@@ -116,6 +116,20 @@ camperRouter.get("/:chargeId/:sessionId", async (req, res) => {
   }
 });
 
+// ROLES: unprotected
+/* On successful payment, mark camper as paid */
+camperRouter.post("/confirm-payment/:chargeId", async (req, res) => {
+  const { chargeId } = req.params;
+  try {
+    const camper = await camperService.confirmCamperPayment(
+      (chargeId as unknown) as string,
```
dhmi19, line 125: why does `chargeId as string` not just work?

dhmi19 (approved):
The code looks good but I had some questions about the flow here:

1. This endpoint is called at the end when the payment and db entity creation is done right?
2. What does it mean to access the checkout session after the payment process has already completed? 
3. Did we figure out where the chargeId is retrieved from? Why does the stripeAPI retrieve method ask for the chargeId?

e-wai:
> 1. This endpoint is called at the end when the payment and db entity creation is done right?
Yep

> 2. What does it mean to access the checkout session after the payment process has already completed?
Description in https://github.com/uwblueprint/focus-on-nature/pull/236 helps a bit - we cache the checkout session data in session storage, and then try to recover it once the payment redirects

> 3. Did we figure out where the chargeId is retrieved from? Why does the stripeAPI retrieve method ask for the charged?
So I use the checkout session ID as the chargeid -- that's what's needed to fetch any given session https://stripe.com/docs/api/checkout/sessions/retrieve

PR #234, opened 2022-12-30
[2/n] Modify registerCamper endpoints with Stripe creation

e-wai:
## Notion ticket link

## Implementation description
* 

## Steps to test
1.

* 

## Checklist

### DIFF
--- a/backend/typescript/services/implementations/camperService.ts
+++ b/backend/typescript/services/implementations/camperService.ts
@@ -120,6 +120,15 @@ class CamperService implements ICamperService {
         );
       }
 
+      // Cannot register for sessions in progress
+      if (
+        !sessionsToRegister.every((cs) =>
+          cs.dates.every((csDate) => csDate.getTime() > Date.now()),
+        )
+      ) {
+        throw new Error("Cannot register for sessions in progress or finished");
+      }
+
       const camp = await MgCamp.findById(
         sessionsToRegister[0].camp,
         {},
@@ -131,7 +140,12 @@ class CamperService implements ICamperService {
         );
       }
 
-      // Ensure all sessions passed in belong to the same camp
+      if (!camp.active) {
+        throw new Error(
+          `Camp associated with camp sessions is a draft camp, must publish camp to add campers`,
+        );
+      }
+
       if (!sessionsToRegister.every((cs) => cs.camp.toString() === camp.id)) {
         throw new Error("All sessions must belong to the same camp");
       }
@@ -321,7 +335,7 @@ class CamperService implements ICamperService {
       await session.commitTransaction();
     } catch (error: unknown) {
       Logger.error(
-        `Failed to create camper. Reason: ${getErrorMessage(error)}`,
+        `Failed to create camper(s). Reason: ${getErrorMessage(error)}`,
       );
       await session.abortTransaction();
       throw error;
--- a/backend/typescript/services/implementations/camperService.ts
+++ b/backend/typescript/services/implementations/camperService.ts
@@ -14,13 +14,18 @@ import {
   WaitlistedCamperDTO,
   UpdateCamperDTO,
   CamperCharges,
+  CampRegistrationDTO,
 } from "../../types";
 import { getErrorMessage } from "../../utilities/errorUtils";
 import logger from "../../utilities/logger";
 import IEmailService from "../interfaces/emailService";
 import nodemailerConfig from "../../nodemailer.config";
 import EmailService from "./emailService";
-import { retrieveStripeCheckoutSession } from "../../utilities/stripeUtils";
+import {
+  createStripeCheckoutSession,
+  createStripeLineItems,
+  retrieveStripeCheckoutSession,
+} from "../../utilities/stripeUtils";
 
 const Logger = logger(__filename);
 const emailService: IEmailService = new EmailService(nodemailerConfig);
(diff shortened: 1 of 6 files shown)


backend/typescript/utilities/stripeUtils.ts
```diff
@@ -90,6 +92,83 @@ export async function createStripePrice(
...
+): Stripe.Checkout.SessionCreateParams.LineItem[] => {
+  return sessionsToRegister.flatMap((campSession) => {
+    const priceItems = [
+      { price: campSession.campPriceId, quantity: campers.length },
+    ];
+
+    // every camper belongs to every session
+    // all same EDLP price
+    // campers have different EDLP dates, which may belong to different sessions
+    // we want to isolate EDLP per session
```
e-wai, line 109: todo remove this

backend/typescript/utilities/stripeUtils.ts
```diff
@@ -90,6 +92,83 @@ export async function createStripePrice(
...
+          const lpDates = campSession.dates.filter((date) =>
+            camperLPDates.has(date.toString()),
+          );
+          camperLateTotal += lpDates.length;
+        }
+
+        return [earlyTotal + camperEarlyTotal, lateTotal + camperLateTotal];
+      },
+      [0, 0],
+    );
```
e-wai, line 136: this is wrong -- identifies # of days, not # of EDLP slots. correct in later PR

backend/typescript/services/implementations/camperService.ts
```diff
@@ -97,10 +103,12 @@ class CamperService implements ICamperService {
         {},
         { session },
       );
+
       // Ensure all session ids passed in are found in the db
       if (sessionsToRegister.length !== campSessions.length) {
         throw new Error("Not all session ids passed in were found");
       }
+
```
dhmi19, line 111: I think I forgot to check if the camp is a published camp which is not completed. Could you add the camp status check (there should be a get status function somewhere in utils- also idk why a camp doesn't just have a status but here we are)

backend/typescript/services/implementations/camperService.ts
```diff
@@ -144,10 +183,10 @@ class CamperService implements ICamperService {
             latePickup: camper.latePickup,
             specialNeeds: camper.specialNeeds,
             contacts: camper.contacts,
-            registrationDate: camper.registrationDate,
-            hasPaid: camper.hasPaid,
+            registrationDate: new Date().toString(),
+            hasPaid: false,
             formResponses: camper.formResponses,
-            chargeId: camper.chargeId,
+            chargeId: createStripeCheckoutSessionResponse.id,
```
dhmi19, line 189: ahh gotcha, ignore my comment in the last PR

backend/typescript/types.ts
```diff
@@ -85,7 +90,7 @@ export type CamperCSVInfoDTO = {
   "Requires Early Drop-off": string;
   "Requires Late Pick-up": string;
   // eslint-disable-next-line
-  "Allergies": string;
+  Allergies: string;
```
dhmi19, line 93: Why are we making this change?

backend/typescript/utilities/stripeUtils.ts
dhmi19: If you keep this logic, I think it's worth having explanation comments here

backend/typescript/utilities/stripeUtils.ts
dhmi19: If we only care about the number of EDLP slots, then we can always just compare against the camp start and end time right? All sessions in the camp will start and end at the same time. So we might not care about which session the EDLP slot belongs to if we can just use the start and end time to calculate the number of slots for ED and LP. 

Also the camper model has 2 separate array fields for ED and LP so you could use those to identify if you need start or end time

backend/typescript/services/implementations/camperService.ts
```diff
@@ -128,8 +136,39 @@ class CamperService implements ICamperService {
...
+        !campers.every(
+          (camper) =>
+            camper.age >= camp.ageLower && camper.age <= camp.ageUpper,
+        )
+      ) {
+        throw new Error("Not all campers meet the age requirements");
+      }
+
+      // Add the total charges for the campers:
+      const lineItems = createStripeLineItems(
```
dhmi19, line 150: do we need these line items later at some point or can we access them through the checkout session?

dhmi19 (approved):
Couple of comments but the logic seems ok to me

backend/typescript/utilities/stripeUtils.ts
dhmi19: nvm I think you did that in the next PR

e-wai: yeah I need another PR to fix a lot of the EDLP logic

backend/typescript/services/implementations/camperService.ts
e-wai: Not sure we'll need them elsewhere?

backend/typescript/utilities/stripeUtils.ts
e-wai: yeah this will disappear soon

backend/typescript/types.ts
e-wai: This was an automatic linter change for me

PR #239, opened 2023-01-01
fix: Form Responses empty object bug in Registrant Experience

twilkhoo:
## Notion ticket link

[Registrant Experience: Camper formQuestions become empty objects from maps](https://www.notion.so/uwblueprintexecs/Registrant-Experience-Camper-formQuestions-become-empty-objects-from-maps-2f7fb75e14a04c189396be841b6ba019)

## Implementation description
* The problem was that in making a deep copy (check changed files), the JSON.stringify/parse methods would convert the formResponses maps into empty objects since maps aren't serializable, and then using the .get method on the empty objects would fail (since they're only supposed to work on maps).
* To fix this, I added a for loop to copy over the maps from `Campers` after the `newCampers` array is made.

## Steps to test
1. Register for any camp i.e. http://localhost:3000/register/camp/63139c7bc3d7b55b44a01531
2. Fill out step 1
3. Fill out step 2
4. Go back to step 1 and try to update the camper information, if it doesn't crash it works (currently this causes it to crash due to the problem above).

* Code correctness.

## Checklist

### DIFF
--- a/frontend/src/components/pages/RegistrantExperience/RegistrationSteps/AdditionalInfo/QuestionCards/CamperQuestionsCard.tsx
+++ b/frontend/src/components/pages/RegistrantExperience/RegistrationSteps/AdditionalInfo/QuestionCards/CamperQuestionsCard.tsx
@@ -31,7 +31,7 @@ const CamperQuestionsCard = ({
         <Wrap width="100%" px="20px" justify="space-between">
           {campSpecificFormQuestions.map((question) => (
             <WrapItem
-              key={`additional_info_question_${question}`}
+              key={`additional_info_question_${question.id}`}
               width={{ sm: "100%", md: mdWrapWidth }}
               px="20px"
               py="12px"
--- a/frontend/src/components/pages/RegistrantExperience/RegistrationSteps/PersonalInfo/personalInfoReducer.tsx
+++ b/frontend/src/components/pages/RegistrantExperience/RegistrationSteps/PersonalInfo/personalInfoReducer.tsx
@@ -19,6 +19,11 @@ export const CamperReducer = (
     const newCampers: RegistrantExperienceCamper[] = JSON.parse(
       JSON.stringify(campers),
     ); // Deep Copy
+
+    for (let i = 0; i < campers.length; i += 1) {
+      newCampers[i].formResponses = campers[i].formResponses;
+    } // Copy the formResponses map
+
     switch (action.type) {
       case PersonalInfoActions.ADD_CAMPER: {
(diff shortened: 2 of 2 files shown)


frontend/src/components/pages/RegistrantExperience/RegistrationSteps/AdditionalInfo/QuestionCards/CamperQuestionsCard.tsx
```diff
@@ -31,7 +31,7 @@ const CamperQuestionsCard = ({
         <Wrap width="100%" px="20px" justify="space-between">
           {campSpecificFormQuestions.map((question) => (
             <WrapItem
-              key={`additional_info_question_${question}`}
+              key={`additional_info_question_${question.id}`}
```
twilkhoo, line 34: With template strings, ${object} won't display the entire object data so we end up having duplicate keys that all say "additional_info_question_[object]" or something of that sort, so we need the id.

frontend/src/components/pages/RegistrantExperience/RegistrationSteps/PersonalInfo/personalInfoReducer.tsx
```diff
@@ -19,6 +19,11 @@ export const CamperReducer = (
     const newCampers: RegistrantExperienceCamper[] = JSON.parse(
       JSON.stringify(campers),
     ); // Deep Copy
+
```
dhmi19, line 22: Do we also need this for the additionalInfoReducer?

frontend/src/components/pages/RegistrantExperience/RegistrationSteps/AdditionalInfo/QuestionCards/CamperQuestionsCard.tsx
dhmi19: yeah good catch. I also made the same change on my feature branch but I'll just update it to match this change

dhmi19 (approved):
ty for this! Just a left a comment to see if we need to make a similar change in the other reducer that deals with formResponses

frontend/src/components/pages/RegistrantExperience/RegistrationSteps/PersonalInfo/personalInfoReducer.tsx
twilkhoo: From my testing no we don't, since the other reducer doesn't use JSON.parse/stringify in this same way.

PR #247, opened 2023-01-13
Move campers modal

TW1234567:
## Notion ticket link

[Camp Overview: Move Camper(s) to different session](https://www.notion.so/uwblueprintexecs/Camp-Overview-Move-Camper-s-to-different-session-16ca401098364231a674e581ea836899)

## Implementation description
* Updated `updateCamperByIds` to work with editing session information
* Edited `EditCamperInfoField` to accept all optional fields (This is done because the type is only used to `updateCamperByIds` and I felt that it's fine to not enforce updating specific fields all the time on fn call)

## Steps to test
1. I had some confusion over how the dates were stored for camp sessions. The date object holds a list of 3 date strings. I'm not sure what the 3 mean. Start date?  final registration date?  end date? (In that order?)

* Main frontend components are `MoveModal` and `SelectModal`
* Main backend change was to `updateCamperByIds`
[Untitled_ Jan 12, 2023 11_30 PM.webm](https://user-images.githubusercontent.com/117593511/212237980-fe3c61da-014e-4b38-8c69-595dd776ad9b.webm)

## Checklist

### DIFF
--- a/backend/typescript/services/implementations/camperService.ts
+++ b/backend/typescript/services/implementations/camperService.ts
@@ -718,7 +718,13 @@ class CamperService implements ICamperService {
         }
 
         const timeNow = new Date();
-        const newCampSessionDate = newCampSession.dates[0];
+        // Find the earliest date for the camp session
+        const earliestDate = newCampSession.dates.reduce(
+          (earliest, date) => (earliest < date ? earliest : date),
+          new Date(Infinity),
+        );
+
+        const newCampSessionDate = earliestDate;
         if (timeNow > newCampSessionDate) {
           throw new Error(
             `Error: you can only move campers to future sessions`,
--- a/backend/typescript/services/implementations/camperService.ts
+++ b/backend/typescript/services/implementations/camperService.ts
@@ -662,7 +662,6 @@ class CamperService implements ICamperService {
     let oldCampSession: CampSession | null = null;
     let movedCampSession = false;
     let camp: Camp | null;
-    console.log("updat ecmapers called", camperIds, updatedFields)
     try {
       oldCampers = await MgCamper.find({
         _id: {
@@ -718,21 +717,22 @@ class CamperService implements ICamperService {
           );
         }
 
-        const timeNow = new Date()
-        const newCampSessionDate = newCampSession.dates[0]
-        if (timeNow > newCampSessionDate){
-          throw new Error (
+        const timeNow = new Date();
+        const newCampSessionDate = newCampSession.dates[0];
+        if (timeNow > newCampSessionDate) {
+          throw new Error(
             `Error: you can only move campers to future sessions`,
-          )
+          );
         }
 
-        const newCampSessionCapcityRemaining = newCampSession.capacity - newCampSession.campers.length
+        const newCampSessionCapcityRemaining =
+          newCampSession.capacity - newCampSession.campers.length;
         const oldCampSessionOriginalCampers = oldCampSession.campers; // for roll back
 
-        if (oldCampSessionOriginalCampers.length > newCampSessionCapcityRemaining){
-          throw new Error(
-            `Error: camp session to move to is full`
-          )
+        if (
+          oldCampSessionOriginalCampers.length > newCampSessionCapcityRemaining
+        ) {
+          throw new Error(`Error: camp session to move to is full`);
         }
 
         // campers for new camp session
@@ -742,7 +742,7 @@ class CamperService implements ICamperService {
         oldCampSession.campers = oldCampSession.campers.filter(
           (camperId) => !camperIds.includes(camperId.toString()),
         );
-  
+
         // Remove campers from old camp session
         await MgCampSession.updateOne(
           { _id: oldCampSession.id },
--- a/backend/typescript/services/implementations/camperService.ts
+++ b/backend/typescript/services/implementations/camperService.ts
@@ -718,17 +718,42 @@ class CamperService implements ICamperService {
           );
         }
 
+        const timeNow = new Date()
+        const newCampSessionDate = newCampSession.dates[0]
+        if (timeNow > newCampSessionDate){
+          throw new Error (
+            `Error: you can only move campers to future sessions`,
+          )
+        }
+
+        const newCampSessionCapcityRemaining = newCampSession.capacity - newCampSession.campers.length
         const oldCampSessionOriginalCampers = oldCampSession.campers; // for roll back
 
-        if (newCampSession) {
-          // campers for new camp session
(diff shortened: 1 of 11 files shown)


frontend/src/types/CamperTypes.ts
```diff
@@ -51,7 +51,7 @@ export type CreateCamperDTO = Omit<
 };
 
 export type EditCamperInfoFields = Omit<
-  Camper,
+  Partial<Camper>,
```
twilkhoo, line 54: Can you explain why this was needed? Don't fully understand

frontend/src/components/pages/CampOverview/RemoveCamperModal/index.tsx
```diff
@@ -74,7 +74,8 @@ const RemoveCamperModal = ({
       setRetrievedCampers([]);
       setCampersToBeDeleted(new Set<Camper>());
     };
-  }, [camper, toast]);
+    // eslint-disable-next-line
+  }, [camper]);
```
twilkhoo, line 78: Does it cause a problem if we keep toast in the dependency array? If not I suggest we keep it here.

backend/typescript/types.ts
```diff
@@ -241,7 +241,9 @@ export type UpdateCamperDTO = Omit<
   | "latePickup"
   | "contacts"
   | "optionalClauses"
->;
+  | "campSession"
+> &
+  Partial<Pick<CamperDTO, "campSession">>;
```
twilkhoo, line 246: Similar comment to below, why is this needed?

backend/typescript/services/implementations/camperService.ts
```diff
@@ -718,18 +717,50 @@ class CamperService implements ICamperService {
...
+        const oldCampSessionOriginalCampers = oldCampSession.campers; // for roll back
+
+        if (
+          oldCampSessionOriginalCampers.length > newCampSessionCapcityRemaining
+        ) {
+          throw new Error(`Error: camp session to move to is full`);
+        }
+
+        // campers for new camp session
+        newCampSession.campers = newCampSession.campers.concat(oldCampers);
```
twilkhoo, line 745: Looks like you removed the if clause around this and the `oldCampSession.campers...`  lines, was this a neccesary change? If not, we could put it back but just verify once.

twilkhoo: Nvm, I see we still `have if (!newCampSession)`, which makes the old if statement unneccesary, it's good you removed it.

frontend/src/components/pages/CampOverview/MoveCamperModal/MoveModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+  selectedCampSession,
+  setSelectedCampSession,
+  deselectAndClose,
+  moveCampers,
+}: PageTwoProps): JSX.Element => {
+  return (
+    <>
+      <ModalHeader>Move {campersToBeMoved.length} Camper(s)</ModalHeader>
+      <ModalBody>
+        <p style={{ marginBottom: "20px" }}>
```
twilkhoo, line 36: Can you change this to a `<Text>` or remove the p tag in some other way?

frontend/src/components/pages/CampOverview/MoveCamperModal/MoveModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+                    year: "numeric",
+                  })}{" "}
+                  ({campSessionsItem.capacity - campSessionsItem.campers.length}{" "}
+                  spots open)
+                </option>
+              );
+            })}
+          </Select>
+        ) : (
+          <p>Error loading camp sessions.</p>
```
twilkhoo, line 91: Same here, remove the <p> and opt for a <Text>, if possible.

twilkhoo (commented):
Really great work! I left some comments (mostly questions, some requests to edit some components). I also added a couple of my own (minor) changes- fixing the cancel button invoke and removing the X at the top left as in the designs. Thanks for getting this done!

twilkhoo (changes requested):
Requesting the changes mentioned above.

twilkhoo:
One more thing I noticed- at http://localhost:3000/admin/camp/6396319f01dd5e987a03b516, I can't move campers to session 1. Could you check this out too? Might be some bug, but lmk if it makes sense.

frontend/src/types/CamperTypes.ts
TW1234567: Yep! Firstly, from my POV, `EditCamperInfoFields` is used to store fields to update for camper(s). For example, filling in `EditCamperInfoField.hasPaid` means to say that. "hey, we want these camper(s) to have their hasPaid set to whatever". Previously, `EditCamperInfoFields` was forced to have some fields (exp, firstname and lastname) be required. I wanted to use the `EditCamperInfoFields` type for moving camper(s) by setting `EditCamperInfoFields.sessionId`. In order to do so, I needed all of those required fields to be optional.

frontend/src/components/pages/CampOverview/RemoveCamperModal/index.tsx
TW1234567: Yeah, my mistake. Will put toast into dependency array.

backend/typescript/types.ts
TW1234567: Good point. I'll remove lines 244, 245, and 246. the reason I included it was because I though that having `UpdateCamperDTO.campSession` be required might lead to runtime errors when `UpdateCamperDTO.campSession` isn't provided (as in the case of edit camper modal). But I guess it's just filled in to be undefined by some other code.

frontend/src/components/pages/CampOverview/MoveCamperModal/MoveModal.tsx
TW1234567: Yep! Will do

frontend/src/components/pages/CampOverview/MoveCamperModal/MoveModal.tsx
TW1234567: Yep!

frontend/src/types/CamperTypes.ts
twilkhoo: Ah okay that makes sense, it's good that you're not making new types. My only concern is that by making camper Partial, basically everything else we've defined is optional too (maybe dangerous?), but seeing as we're nearing the end of the project, we should be alright.

twilkhoo (approved):
LGTM

PR #249, opened 2023-01-15
Fix: camp creation validator logic

e-wai:
Add states to double check for non-negative, lower < upper, etc.

## Notion ticket link

## Implementation description
* 

## Steps to test
1.

* 

## Checklist

### DIFF
--- a/frontend/src/components/pages/CampCreation/index.tsx
+++ b/frontend/src/components/pages/CampCreation/index.tsx
@@ -59,27 +59,43 @@ const CampCreationPage = (): React.ReactElement => {
 
   let isCampDetailsFilled = false;
 
-  const startTimeBeforeEndTime = useMemo(() => {
-    try {
-      const [startHours, startMinutes] = startTime
-        .split(":")
-        .map((num) => parseInt(num, 10));
-      const [endHours, endMinutes] = endTime
-        .split(":")
-        .map((num) => parseInt(num, 10));
+  const useValidateTimeOrder = (start: string, end: string): boolean => {
+    return useMemo(() => {
+      try {
+        const [startHours, startMinutes] = start
+          .split(":")
+          .map((num) => parseInt(num, 10));
+        const [endHours, endMinutes] = end
+          .split(":")
+          .map((num) => parseInt(num, 10));
 
-      if (!(startHours && startMinutes && endHours && endMinutes)) {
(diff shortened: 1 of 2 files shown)


twilkhoo (dismissed):
LGTM, everything works on my end, just wondering if a camp's fee can ever be $0? Your implementation requires camp fees to be >=$1, so just curious about that.

twilkhoo (commented):
One more question- do we want error validation for EDLP similar for start and end time? I.e. if early dropoff time is later than late pickup time, that should also be an error right?

frontend/src/components/pages/CampCreation/index.tsx
```diff
@@ -59,18 +59,42 @@ const CampCreationPage = (): React.ReactElement => {
...
+  const startTimeBeforeEndTime = useMemo(() => {
+    try {
+      const [startHours, startMinutes] = startTime
+        .split(":")
+        .map((num) => parseInt(num, 10));
+      const [endHours, endMinutes] = endTime
+        .split(":")
+        .map((num) => parseInt(num, 10));
+
+      if (!(startHours && startMinutes && endHours && endMinutes)) {
```
twilkhoo, line 71: Replace this if condition with:
```
if (Number.isNaN(startHours) || Number.isNaN(endHours)) {
  return true;
}
```
Right now, if you have times with 0, these will also pass the validation due to falsey value logic. Also, there's no need to check for startMinutes in the if condition, if either hours or minutes are not present, then hours will return NaN.

twilkhoo (approved):
LGTM

frontend/src/components/pages/CampCreation/index.tsx
```diff
@@ -59,18 +59,76 @@ const CampCreationPage = (): React.ReactElement => {
...
+          .map((num) => parseInt(num, 10));
+        const [endHours, endMinutes] = end
+          .split(":")
+          .map((num) => parseInt(num, 10));
+
+        if (Number.isNaN(startHours) || Number.isNaN(endHours)) {
+          return true;
+        }
+
+        if (canBeEqual) {
```
twilkhoo, line 80: Checked with Andy, EDLP is allowed to have the same times as the camp's start and end time, so I added this `canBeEqual` check.

e-wai:
beauty

PR #257, opened 2023-01-23
Fix: Render dates correctly

twilkhoo:
## Notion ticket link

[[Registrant Experience] Dates to select session are incorrect](https://www.notion.so/uwblueprintexecs/Registrant-Experience-Dates-to-select-session-are-incorrect-ee341f9de21d421eb8be0bc903ea4843)

## Implementation description
* This ticket accounts for all problems in rendering dates
* The biggest issue that I encountered in a lot of places is that we do not sort the dates. The dates in the order which they appear in an object are not sorted, so simply taking the first and last elements in the dates array is not sufficient, that's why we display strange date strings (shown below).

## Steps to test
1. Navigate around the entire site (literally everything) and ensure all dates make sense- they should comply with the actual info of the camp object, and the order should also be correct. Let me know if I missed anything. Here are the 5 places where I made changes:

**Before:**
image
image
image
image
image

**After:**
image
image
image
image
image

* Correctness, and also make sure I covered every case of incorrect date strings.

## Checklist

### DIFF
--- a/frontend/src/components/pages/CampsList/CampsTable.tsx
+++ b/frontend/src/components/pages/CampsList/CampsTable.tsx
@@ -1,5 +1,4 @@
 import React, { Dispatch, SetStateAction } from "react";
-
 import { useHistory } from "react-router-dom";
 import { SearchIcon } from "@chakra-ui/icons";
 import { FaEllipsisV } from "react-icons/fa";
@@ -195,10 +194,7 @@ const CampsTable = ({
                 }}
               >
                 {camp.campSessions.length > 0
-                  ? getFormattedCampDateRange(
-                      camp.campSessions[0].dates,
-                      camp.campSessions[camp.campSessions.length - 1].dates,
-                    )
+                  ? getFormattedCampDateRange(camp.campSessions)
                   : ""}
               </Td>
               <Td
--- a/frontend/src/components/pages/RegistrantExperience/SessionSelection/SessionCard.tsx
+++ b/frontend/src/components/pages/RegistrantExperience/SessionSelection/SessionCard.tsx
@@ -7,7 +7,7 @@ import {
   sessionCardDatesTextStyles,
   sessionCardDetailsTextStyles,
 } from "./SessionSelectionStyles";
-import { getMeridianTime } from "../../../../utils/CampUtils";
+import { getMeridianTime, sortDateStrings } from "../../../../utils/CampUtils";
 
 const WAITLISTED_COLOR = "red.100";
 const AVAILABLE_COLOR = "green.100";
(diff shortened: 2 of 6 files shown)


frontend/src/components/pages/CampsList/CampsTable.tsx
```diff
@@ -195,10 +194,7 @@ const CampsTable = ({
                 }}
               >
                 {camp.campSessions.length > 0
-                  ? getFormattedCampDateRange(
-                      camp.campSessions[0].dates,
-                      camp.campSessions[camp.campSessions.length - 1].dates,
-                    )
+                  ? getFormattedCampDateRange(camp.campSessions)
```
twilkhoo, line 197: I edited the function to now take all sessions of a camp, and then sort them. I don't think we guarantee that the order of sessions as they appear in the object indicate order of what dates they are held on (especially considering overlapping sessions) so now I'm taking all sessions and sorting them all at once.

frontend/src/utils/CampUtils.ts
```diff
@@ -56,19 +56,42 @@ export const getCampStatus = (camp: CampResponse): CampStatus => {
...
+  dateNumToStrArr.forEach((date) => {
+    sortedDates.push(date[1]);
+  });
+
+  return sortedDates;
+};
+
+export const getFormattedDateString = (
+  dates: string[],
+  requireSorted = false,
```
twilkhoo, line 78: Added a requireSorted parameter with default value false. I don't think we ever need formatted date strings without sorting them, but if we did want the option, I didn't want to change/break anything else we had, thus the default value of false (and set it to true to get sorted dates).

frontend/src/components/pages/RegistrantExperience/SessionSelection/SessionCard.tsx
```diff
@@ -27,13 +27,14 @@ const SessionCardDetails = ({
...
 }: SessionCardDetailsProps): React.ReactElement => {
   const getCampSessionDates = () => {
-    const startDate = new Date(campSession.dates[0]);
-    const endDate = new Date(campSession.dates[campSession.dates.length - 1]);
+    const datesToUse = sortDateStrings(campSession.dates);
+    const startDate = new Date(datesToUse[0]);
+    const endDate = new Date(datesToUse[campSession.dates.length - 1]);
     return `${startDate.toLocaleString("default", {
       month: "short",
-    })} ${startDate.getDay()} - ${endDate.toLocaleString("default", {
```
twilkhoo, line 34: Funny little bug, getDay() returns day of the week, not the day of the month. That's what getDate() is for.

PR #258, opened 2023-01-23
Fix: Display EDLP tag applicable to a session

twilkhoo:
## Notion ticket link

[[Registrant Experience] Dates to select session are incorrect](https://www.notion.so/uwblueprintexecs/Registrant-Experience-Dates-to-select-session-are-incorrect-ee341f9de21d421eb8be0bc903ea4843)

## Implementation description
* Passed in a particular session as a prop
* Using the session's dates, I can conditionally render the EDLP tags if the camper's selected EDLP dates are within the particular session's range

## Steps to test
1. Go to http://localhost:3000/admin/camp/6396319f01dd5e987a03b516
2. Ensure that the campers that have EDLP only display the purple EDLP tags if the EDLP date is in range.

**Before:**
image

**After:**
image

* Correctness, functionality.

## Checklist

### DIFF
--- a/frontend/src/components/pages/CampOverview/EditCamperModal/index.tsx
+++ b/frontend/src/components/pages/CampOverview/EditCamperModal/index.tsx
@@ -4,7 +4,6 @@ import {
   ModalOverlay,
   ModalContent,
   ModalHeader,
-  ModalCloseButton,
   Modal,
   ModalBody,
   ModalFooter,
@@ -131,7 +130,6 @@ const EditCamperModal = ({
         <ModalHeader>
           Edit {camper.firstName} {camper.lastName}&apos;s Information
         </ModalHeader>
-        <ModalCloseButton />
         <ModalBody overflowY="scroll">
           <EditCamperModalForm
             formQuestions={formQuestions}
--- a/frontend/src/components/pages/CampOverview/ViewCamperModal/PickupDropoffTableRow.tsx
+++ b/frontend/src/components/pages/CampOverview/ViewCamperModal/PickupDropoffTableRow.tsx
@@ -54,7 +54,7 @@ const PickupDropoffTime = ({
   return (
     <Td pl={0} pt={2} pb={0} color={color}>
       <HStack>
-        {pickup ? <SunsetIcon fill={color} /> : <SunriseIcon fill={color} />}
+        {pickup ? <SunriseIcon fill={color} /> : <SunsetIcon fill={color} />}
         <Text textStyle="bodyRegular" color={color}>
           {time}
         </Text>
--- a/frontend/src/components/pages/CampOverview/ViewCamperModal/PickupDropoffTableRow.tsx
+++ b/frontend/src/components/pages/CampOverview/ViewCamperModal/PickupDropoffTableRow.tsx
@@ -54,7 +54,7 @@ const PickupDropoffTime = ({
   return (
     <Td pl={0} pt={2} pb={0} color={color}>
       <HStack>
-        {pickup ? <SunriseIcon fill={color} /> : <SunsetIcon fill={color} />}
+        {pickup ? <SunsetIcon fill={color} /> : <SunriseIcon fill={color} />}
         <Text textStyle="bodyRegular" color={color}>
           {time}
         </Text>
--- a/frontend/src/utils/CampUtils.ts
+++ b/frontend/src/utils/CampUtils.ts
@@ -490,19 +490,12 @@ export const checkDatesInRange = (
 ): boolean => {
   if (!datesToCheck || datesToCheck.length === 0) return false;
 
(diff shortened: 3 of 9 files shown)


frontend/src/components/pages/CampOverview/EditCamperModal/index.tsx
```diff
@@ -131,7 +130,6 @@ const EditCamperModal = ({
         <ModalHeader>
           Edit {camper.firstName} {camper.lastName}&apos;s Information
         </ModalHeader>
-        <ModalCloseButton />
```
rchang0501, line 134: just wanted to confirm the reason for removing this button does it just not match the designs?

frontend/src/components/pages/CampOverview/ViewCamperModal/PickupDropoffTableRow.tsx
```diff
@@ -54,7 +54,7 @@ const PickupDropoffTime = ({
   return (
     <Td pl={0} pt={2} pb={0} color={color}>
       <HStack>
-        {pickup ? <SunriseIcon fill={color} /> : <SunsetIcon fill={color} />}
```
rchang0501, line 57: we need to revert this, the pickup icon is supposed to be the sunrise icon not the sunset one (I know it's kinda confusing)

rchang0501: Expected look:

Screen Shot 2023-01-28 at 2 12 05 PM

frontend/src/utils/CampUtils.ts
```diff
@@ -483,3 +483,26 @@ export const getFormattedCampDateRange = (camp: CampResponse): string => {
     ? `${format(firstCampDate, "PP")} - ${format(lastCampDate, "PP")}`
     : "";
 };
+
+export const checkDatesInRange = (
```
rchang0501, line 487: I think we can simplify this function a bit - smthn like this: 

```
export const checkDatesInRange = (
  rangeDates: string[],
  datesToCheck?: Date[],
): boolean => {
  if (!datesToCheck || datesToCheck.length === 0) return false;

  const minDate = Math.min(...rangeDates.map((date) => Date.parse(date)));
  const maxDate = Math.max(...rangeDates.map((date) => Date.parse(date)));

  return datesToCheck.some(
    (date: Date) =>
      new Date(date).getTime() >= minDate &&
      new Date(date).getTime() <= maxDate,
  );
};
```

frontend/src/components/pages/CampOverview/ViewCamperModal/PickupDropoffTableRow.tsx
```diff
@@ -54,7 +54,7 @@ const PickupDropoffTime = ({
...
+        {pickup ? <SunsetIcon fill={color} /> : <SunriseIcon fill={color} />}
```
rchang0501, line 57: I believe need to revert this - pickup should be the sunrise button (I know it's kinda confusing)

Expected look: 
Screen Shot 2023-01-28 at 2 12 05 PM

Design ref:
Screen Shot 2023-01-28 at 2 13 30 PM

rchang0501 (changes requested):
overall looks good! 

left 2 comments and one clarification question - lmk if you have any questions!

frontend/src/components/pages/CampOverview/ViewCamperModal/PickupDropoffTableRow.tsx
twilkhoo: Sounds good, I'll revert it but I think I'll make a lower priority ticket just going over the inconsistencies of these icons. 
image
Here, early dropoff is sunrise and late pickup is sunset (which IMO, makes a lot more sense, but we can verify with the designers)

frontend/src/components/pages/CampOverview/EditCamperModal/index.tsx
twilkhoo: Yeah, the modals don't have the close button (consistency).

frontend/src/utils/CampUtils.ts
twilkhoo: Works like a charm, thanks!

frontend/src/components/pages/CampOverview/ViewCamperModal/PickupDropoffTableRow.tsx
rchang0501: yep sounds good, i totally thought it was the other way around initially as well but I feel like if we are gonna make a change we should choose the easier one to change (ie the most consistent set we have rn)

rchang0501 (approved):
lgtm!

PR #265, opened 2023-02-09
[Registrant Experience] Prevent user from proceeding if anything on review page is currently being edited

twilkhoo:
## Notion ticket link

[[Registrant Experience] Prevent user from proceeding if anything on review page is currently being edited](https://www.notion.so/uwblueprintexecs/Registrant-Experience-Prevent-user-from-proceeding-if-anything-on-review-page-is-currently-being-ed-10bb5b79972647989cc3d1b3f9716e66?pvs=4)

## Implementation description
* Added another piece of state, called `isEditing`, which is a number that counts the amount of components being edited (0 means nothing is being edited)
* This does in some sense create another source of truth for whether something is editing or not (global vs local state for each edit card), so I made sure to be careful and increment/decrement only when saving or canceling, so registrant's never get stuck on the review page.

## Steps to test
1. Go to http://localhost:3000/register/camp/6396319f01dd5e987a03b516 and try moving forward or backward while editing (you shouldn't be able to).

https://user-images.githubusercontent.com/30396273/217713242-368f0eaf-3c59-4996-bb82-191a6e9e909a.mp4

* Please test extensively, it's important that this has no issues otherwise registration would be impossible.

## Checklist

### DIFF
--- a/frontend/src/components/pages/RegistrantExperience/RegistrationSteps/RegistrationFooter.tsx
+++ b/frontend/src/components/pages/RegistrantExperience/RegistrationSteps/RegistrationFooter.tsx
@@ -10,6 +10,7 @@ export type RegistrationFooterProps = {
   handleStepNavigation: (stepsToMove: number) => void;
   isPaymentSummary: boolean;
   isWaitlistRegistration: boolean;
+  isEditing: number;
 };
 
 const RegistrationFooter = ({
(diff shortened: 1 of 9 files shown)


rchang0501 (approved):
Tested pretty extensively and didn't find any issues! Don't have any code comments either since it was primarily state management that made sense

PR #267, opened 2023-02-10
Resolve uncaught error when viewing camp and fix coordinator/counsellor/volunteer editing

twilkhoo:
## Notion ticket link

[Uncaught error when viewing camp](https://www.notion.so/uwblueprintexecs/Uncaught-error-when-viewing-camp-e7dc9d9ca33e4e72abf171da5fee9869?pvs=4)
[Camp Overview: Fix Camp Edit functionality](https://www.notion.so/uwblueprintexecs/Camp-Overview-Fix-Camp-Edit-functionality-f33023f2ca704b7cb8630fda072a857a?pvs=4)

## Implementation description
* Had to change some of the validators.
* PR #181 refactored create and update validators into a single validator, but turns out we can't have absolutely everything common between the two, so I created two validators (create, update) that implement the old common validator with some minor adjustments.
* After resolving the error, I had to change updateCampById, more information is annotated in the files changed.

## Steps to test
1. Ensure the error is gone when viewing a camp.
2. Test literally everything- all parts of the site including registration, camp creation, camp drafting, camp editing, camp viewing, everything else just to be safe. Notably, make sure that the camp coordinator/counsellor/volunteer component works now. This is to make sure we don't have to redo anything, although the components that I changed were not used anywhere else other than when viewing a camp.

* Correctness, functionality.

## Checklist

### DIFF
--- a/backend/typescript/middlewares/validators/campValidators.ts
+++ b/backend/typescript/middlewares/validators/campValidators.ts
@@ -218,8 +218,8 @@ export const createUpdateCampDtoValidator = async (
       const campSession = body.campSessions[i];
       try {
         createUpdateCampSessionDTOValidator(campSession, type);
-      } catch (err: any) {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      } catch (err: any) {
         res.status(400).send(err.message);
       }
     }
--- a/backend/typescript/middlewares/validators/campValidators.ts
+++ b/backend/typescript/middlewares/validators/campValidators.ts
@@ -16,10 +16,15 @@ import {
 } from "./util";
 import { getErrorMessage } from "../../utilities/errorUtils";
 
-const createCampSessionDTOValidator = (campSession: {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  [key: string]: any;
-}): boolean => {
+enum CreateOrUpdate {
+  CREATE = 1,
+  UPDATE = 2,
+}
+
+const createUpdateCampSessionDTOValidator = (
+  campSession: { [key: string]: any }, // eslint-disable-line @typescript-eslint/no-explicit-any
+  type: CreateOrUpdate,
+): boolean => {
   if (!validateArray(campSession.dates, "string")) {
     throw new Error(getApiValidationError("dates", "string", true));
   }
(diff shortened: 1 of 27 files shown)


backend/typescript/middlewares/validators/campValidators.ts
```diff
@@ -235,6 +243,22 @@ export const createUpdateCampDtoValidator = async (
   return next();
 };
 
+export const createCampDtoValidator = async (
```
twilkhoo, line 246: These were the two new validators created (create, update), basically just calling the same validator as before but now passing in the type.

backend/typescript/middlewares/validators/campValidators.ts
```diff
@@ -16,10 +16,15 @@ import {
...
-const createCampSessionDTOValidator = (campSession: {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  [key: string]: any;
-}): boolean => {
+enum CreateOrUpdate {
+  CREATE = 1,
+  UPDATE = 2,
+}
+
+const createUpdateCampSessionDTOValidator = (
```
twilkhoo, line 24: `createUpdateCampValidator` calls `createUpdateCampSessionDTOValidator`. This caused the initial problem of "campers must be empty", this validation doesn't make sense when viewing/updating a camp, only when creating.

backend/typescript/services/implementations/campService.ts
```diff
@@ -516,21 +516,6 @@ class CampService implements ICampService {
         throw new Error(`Camp' with campId ${campId} not found.`);
       }
 
-      // Update the campSessions by deleting all current sessions and creating new sessions
```
twilkhoo, line 519: After fixing that problem above regarding the validators, I noticed that campers were being deleted every time we viewed (updated) a camp. This is because of this part of the code which I deleted- shouldn't be a problem anywhere else since we don't use updateCampById for anything other than when we view a camp.

rchang0501 (changes requested):
I tested pretty extensively and everything looks good except for updating camp sessions. Lmk if you have any questions!

backend/typescript/services/implementations/campService.ts
twilkhoo: I'm glad you caught this 😅 you're right- it is a good idea to seperate the logic of the two, so I noticed that we were calling `editCampById` both when on the camp overview (since the volunteer/coordinator editing component would call this), and when we saved or published a draft from the footer (which is what you caught). For the overview we can't have new sessions created, but for the drafting we do need new sessions, and honestly the best way I could find that works was just to create another (yet very similar) end point, both of which call `updateCampById` but now pass in a `createNewSessions` boolean arg which indicates whether we want new sessions or not. Feel free to leave your thoughts on this approach.

rchang0501 (approved):
did another very thorough check - went through pretty much all our flows and couldn't find any issues!

PR #268, opened 2023-02-11
Fix: Manage sessions doesn’t persist

twilkhoo:
## Notion ticket link

[[Management Experience] Manage sessions doesn’t persist](https://www.notion.so/uwblueprintexecs/Management-Experience-Manage-sessions-doesn-t-persist-cec1071f9d9a4cd980973cc6a9e388ab?pvs=4)

## Implementation description
* Changed the format of the patch so that it would actually use the authorization headers
* Changed the validators too to account for this.

## Steps to test
1. Note: This is only going to work after pr #267 is merged.
2. Go to any camp and click the manage sessions button on the right.
3. Changing session capacity should work now, deletion worked anyway.

* Correctness, functionality.

## Checklist

### DIFF
--- a/frontend/src/APIClients/CampsAPIClient.ts
+++ b/frontend/src/APIClients/CampsAPIClient.ts
@@ -105,7 +105,9 @@ const updateCampSessions = async (
   updatedCampSessions: Array<UpdateCampSessionsRequest>,
 ): Promise<Array<CampSessionResponse>> => {
   try {
-    const { data } = await baseAPIClient.patch(`/camp/${campId}/session`, {
+    const { data } = await baseAPIClient({
+      method: "patch",
+      url: `/camp/${campId}/session`,
       headers: {
         Authorization: getBearerToken(),
       },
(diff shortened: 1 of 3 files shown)


frontend/src/APIClients/CampsAPIClient.ts
```diff
@@ -104,7 +104,9 @@ const updateCampSessions = async (
   updatedCampSessions: Array<UpdateCampSessionsRequest>,
 ): Promise<Array<CampSessionResponse>> => {
   try {
-    const { data } = await baseAPIClient.patch(`/camp/${campId}/session`, {
```
twilkhoo, line 107: This is what I meant by "had to change the format", for some reason using client.patch(...) has some strange behaviour in terms of setting/sending headers (authorization in our case), but this other format worked perfectly. I'm still not entirely sure why though, thoughts? I tried to emulate whereever else we used a .patch for anything, but after trying it all, auth headers were still not sending.

rchang0501: i don't see any issue with this!

rchang0501 (approved):
lgtm!

PR #269, opened 2023-02-11
Camp Overview: Sort waitlisted campers by created at date

twilkhoo:
## Notion ticket link

[Camp Overview: Sort waitlisted campers by created at date](https://www.notion.so/uwblueprintexecs/Camp-Overview-Sort-waitlisted-campers-by-created-at-date-d3df8b367eba444a988a58c56fc7f6ce?pvs=4)

## Implementation description
* Added a registrationDate property to waitlisted camper DTOs and schema, and also implemented it whereever needed (pretty much look at whereever waitlisted campers were being used and added the new property there, trying to recreate what we did with the regular camper dtos).

## Steps to test
1. Note: This PR will only work properly after PR #267 is submitted, so I'll draft it until then.
2. Register some campers on a waitlist
3. Ensure that they are sorted in the order you created them in (by registrationDate).

* Correctness, functionality.

## Checklist

### DIFF
--- a/backend/typescript/types.ts
+++ b/backend/typescript/types.ts
@@ -114,6 +114,7 @@ export type WaitlistedCamperDTO = {
   campSession: string;
   status: WaitlistedCamperStatus;
   linkExpiry?: Date;
+  registrationDate: string;
 };
 
 export type CreateUserDTO = Omit<UserDTO, "id">;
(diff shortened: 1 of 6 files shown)


backend/typescript/types.ts
```diff
@@ -114,6 +114,7 @@ export type WaitlistedCamperDTO = {
   campSession: string;
   status: WaitlistedCamperStatus;
   linkExpiry?: Date;
+  registrationDate: string;
```
rchang0501, line 117: is there a reason we're choosing to type this as a string instead of a Date?

rchang0501 (approved):
looks good to me! one thing that may be worth considering is adding the `registrationDate` field to the create waitlisted camper validator `backend/typescript/middlewares/validators/waitlistedCampersValidators.ts`

I don't think we need validation for the field tbh since we're just creating a date object but it may be worth adding to be thorough. Up to you just something I was thinking about

PR #276, opened 2023-05-31
feat: add refund link generation

williamtran10:
## Notion ticket link

[Cancellation/Refund: Generate a unique cancellation link on a per-registration basis](https://www.notion.so/uwblueprintexecs/Cancellation-Refund-Generate-a-unique-cancellation-link-on-a-per-registration-basis-89c51d3fb72a43bbb958f2d4e9da5ef9)

## Implementation description
* Generated unique refund link by generating refund code for each confirmation email sent to parent after they successfully register for a FON camp
* Created endpoint at `/refund/{refundCode}` that returns the refund information associated with the refund code

## Steps to test
1. Go to http://localhost:3000/register/camp/648df427152c24ef5871842a
2. Fill in the required fields and register some campers using an accessible email
3. Once registration is completed, check the email for an email from FON. It should contain a refund link like the one in screenshot below
4. Copy-paste the refundCode (after /refund) to a text doc
5. In Postman, send a GET request to http://localhost:3000/campers/refund/theRefundCodeFromTextDoc
6. Verify response is correct (array of campers registered for that camp and under each camper the sessions they registered for)
Screen Shot 2023-06-17 at 2 00 09 PM

* Check there's a refund link with refund code in the url in the confirmation email
* Check the response from the endpoint is correct

## Checklist

### DIFF
--- a/backend/typescript/services/implementations/camperService.ts
+++ b/backend/typescript/services/implementations/camperService.ts
@@ -53,7 +53,7 @@ class CamperService implements ICamperService {
     let createCamperResponse: CamperDTO[] = [];
     let campersWithRefundCode: Camper[] = [];
     let checkoutSessionUrl = "";
-    let refundCode = "";
+    let code = "";
 
     try {
       // Perform session checks:
@@ -143,8 +143,9 @@ class CamperService implements ICamperService {
       // Create new unused refund code
       do {
         // cryptoRandomString v2.0 takes in an integer for length and outputs alphanumeric random string of length specified
-        refundCode = cryptoRandomString(REFUND_CODE_LENGTH);
-        campersWithRefundCode = await MgCamper.find({ refundCode });
+        code = cryptoRandomString(REFUND_CODE_LENGTH);
+        // eslint-disable-next-line no-await-in-loop
+        campersWithRefundCode = await MgCamper.find({ code });
       } while (campersWithRefundCode && campersWithRefundCode.length > 0);
 
       // Create a camper entity for each session
@@ -171,7 +172,7 @@ class CamperService implements ICamperService {
               earlyDropoff: 0,
               latePickup: 0,
             },
-            refundCode: refundCode,
+            refundCode: code,
             optionalClauses: camper.optionalClauses,
           };
         });
--- a/backend/typescript/services/implementations/camperService.ts
+++ b/backend/typescript/services/implementations/camperService.ts
@@ -1276,6 +1276,8 @@ class CamperService implements ICamperService {
         lastName: "Smith",
         age: 9,
         campName: "Banff Camp 2023",
+        startTime: "17:00",
+        endTime: "20:00",
         instances: [
           {
             id: "63b527fed9c42a8fd5ea55e6",
@@ -1312,6 +1314,8 @@ class CamperService implements ICamperService {
         lastName: "Smith",
         age: 10,
         campName: "Banff Camp 2023",
+        startTime: "17:00",
+        endTime: "20:00",
         instances: [
           {
             id: "63b527fed9c42a8fd5ea55e6",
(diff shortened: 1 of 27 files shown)


backend/typescript/services/implementations/camperService.ts
```diff
@@ -135,6 +140,12 @@ class CamperService implements ICamperService {
 
       checkoutSessionUrl = createStripeCheckoutSessionResponse.url;
 
+      // Create new unused refund code
+      do {
+        refundCode = cryptoRandomString({length: REFUND_CODE_LENGTH, type: 'alphanumeric'});
+        campersWithRefundCode = await MgCamper.find({ checkoutSessionUrl });
```
williamtran10: perhaps change this to `campersWithSameCode` so that its a little more clear whats happening

thywang: thank you, I agree! Have made the change in the latest commit.

backend/typescript/services/implementations/camperService.ts
```diff
@@ -1228,6 +1265,85 @@ class CamperService implements ICamperService {
...
+      if (!camp) {
+        throw new Error(`Camp with ID ${campSession.camp} not found.`);
+      }
+
+      // eslint-disable-next-line no-restricted-syntax
+      for (const camper of campers) {
+        // eslint-disable-next-line no-await-in-loop
+        campSession = await MgCampSession.findById(camper.campSession);
+        key = camper.firstName + camper.lastName;
+        refundCamper = {
```
williamtran10, line 1302: This is a lot of lines to copy data over. Maybe we should just keep all the fields of `Camper` (and change the DTO to reflect that)

backend/typescript/services/implementations/camperService.ts
```diff
@@ -1228,6 +1265,85 @@ class CamperService implements ICamperService {
...
+        );
+      }
+
+      const camp = await MgCamp.findById(campSession.camp);
+      if (!camp) {
+        throw new Error(`Camp with ID ${campSession.camp} not found.`);
+      }
+
+      // eslint-disable-next-line no-restricted-syntax
+      for (const camper of campers) {
```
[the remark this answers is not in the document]
williamtran10: Nevermind, would require changing the rest of the code block too

backend/typescript/services/implementations/camperService.ts
thywang: Agree, that's a great point and idea. I tried this but keeping all the fields of `camper` results in strange formatting of the data. Would propose to keep the copying for easy data handling on the frontend.

backend/typescript/services/implementations/camperService.ts
```diff
@@ -1264,118 +1265,83 @@ class CamperService implements ICamperService {
...
+      const camp = await MgCamp.findById(campSession.camp);
+      if (!camp) {
+        throw new Error(`Camp with ID ${campSession.camp} not found.`);
+      }
+
+      for (let i = 0; i < campers.length; i++) {
+        camper = campers[i];
+        // eslint-disable-next-line no-await-in-loop
+        campSession = await MgCampSession.findById(campers[i].campSession);
+        key = campers[i].firstName + campers[i].lastName;
```
thywang: Currently registration does not allow registering two campers with same name in one checkout session. It only registers the first one and ignores any duplicates.

Using camper's firstName + lastName should be unique key since all campers returned [here](https://github.com/uwblueprint/focus-on-nature/pull/276/commits/09c2133149396a81c75260f0041715c810d6243c#diff-da500917cfe8a24b6558fb1d2419cfca87e7716ee3ab04e08f0a20823dd68821R1279) `const campers: Camper[] = await MgCamper.find({ refundCode });` are registered under same checkout session.

Cannot use camper id since they are different for every camper in campers regardless of their firstName and lastName.

derrick-yyang (approved):
Tested and worked! Code LGTM too

BraydenRoyston (approved):
tested and looks good! merging.

PR #281, opened 2023-06-20
feat: camper selection frontend

derrick-yyang:
## Notion ticket link

[Cancellation-Refund-Camper-Selection-Page](https://www.notion.so/uwblueprintexecs/Cancellation-Refund-Camper-Selection-Page-All-states-9740991233214020bea547b8ae12294e)

## Implementation description
* Implemented the Camper Refund/Cancellation page
* Interacts with backend to pull data with respect to the `refundCode` in the URL and displays the available refund options
**Quick Note:** PR is not finished, still need PM or designer opinion on grouping sessions per name.

## Steps to test
1. Run the app locally and go to `localhost:3000/refund/<any-string>`. The string after refund can be anything for now, the backend will accept any string and will return the same test data
2. Check the validity of the states in each card
3. Check that the cost and numbers are correct
4. Check mobile screen sizes and verify it's what we want according to the figma

* Any styling requests?
* Double check the coding style/design to see if it's reusable/robust enough. I occasionally get errors due to the response from our Backend not being the format our frontend code expects so some functions fail
* **_As per the quick note above, would live an external opinion on grouping sessions under the same name. I feel some more decisions need to be made about this, like: "should all sessions under the same name be refunded all at once or not at all? Or have the option to individually be refunded?", "Do we need to change the formating of camper refund cards?" etc._**

## Checklist

### DIFF
--- a/frontend/src/components/pages/CamperRefundCancellation/CamperRefundFooter.tsx
+++ b/frontend/src/components/pages/CamperRefundCancellation/CamperRefundFooter.tsx
@@ -1,0 +1,34 @@
+import React from "react";
+import { Flex, Button } from "@chakra-ui/react";
+
+const CamperRefundFooter = (props:any): React.ReactElement => {
+
+  return (
+    <Flex
+    color="#FFFFFF"
+    justify="flex-end"
+    position="fixed"
+    width="100%"
+    height="7%"
+    bottom="0"
+    alignItems="flex-end"
+    backgroundColor="#FFFFFF"
+    borderTop="2px solid #EEEFF1"
+    >
+      <Button
+        width="auto"
+        height="48px"
+        variant="primary"
+        background={
+          true ? "primary.green.100" : "primary.green_disabled.100"
+        }
+        mr="50px"
+        mb="12px"
+      >
+        Request refund
+      </Button>
+    </Flex>
+  );
+};
+
+export default CamperRefundFooter;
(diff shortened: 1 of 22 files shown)


NathanSouphanthong (approved):
LGTM

frontend/src/components/pages/CamperRefundCancellation/CamperRefundInfoCard.tsx
```diff
@@ -0,0 +1,222 @@
...
+        dates.push(new Date(date));
+      });
+    });
+
+    // sort the dates and find the first session
+    dates.sort((a, b) => a.getTime() - b.getTime());
+    const firstDate = dates[0];
+    const thirtyDays = 30 * 24 * 60 * 60 * 1000;
+
+    return new Date().getTime() - firstDate.getTime() >= thirtyDays;
```
braydenatnimble, line 93: believe we want to switch this to:
`return (firstData.getTime() - new Date().getTime()) >= thirtyDays`

The current implementation displays both cards as invalid refunds even though the dates are > 30 days away.

braydenatnimble, line 93: nit: while you're here might as well declare this as a const "TODAY" (or similar)
- less magic numbers the better

frontend/src/components/pages/CamperRefundCancellation/CamperRefundFooter.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import { Flex, Button } from "@chakra-ui/react";
+
+const CamperRefundFooter = (props: any): React.ReactElement => {
+  return (
```
braydenatnimble, line 5: Formatting issue: 
Current implementation has the button un-centered. My suggestion:
1) Change alignItems="flex-end" to alignItems="center" on the flex box
2) remove right margin (mr="50px") on the button itself

BraydenRoyston (approved):
Looking good locally with a few minor changes.

Approved once those minor comments are addressed. Good stuff!

PR #295, opened 2023-08-04
Derrick catherine/send data on request refund

derrick-yyang:
## Notion ticket link

No notion link, was follow up changes from RefundCancellation ticket

## Implementation description
* Added functionality where when clicking the request refund button, we will send an HTTP PUT with the DTO to the same endpoint of our GET for refund info
* Added a dummy PUT endpoint that returns an empty array `[]`. I also didn't update the `refunds` state with the response from our PUT endpoint since the endpoint does not exist yet. Thought I'd keep it simple to make it easier to resolve merge conflicts for whoever is working on the actual endpoint
* Added loading state for the button when the API is called

## Steps to test
Can't really test... you'll have to add `console.log`'s in certain places to see if the request is going through

* I added an empty dependency array to `index.tsx` in CamperRefundCancellation since I noticed that our `useEffect` function is being called in an infinite loop. This is because without the empty dependency array, the function in `useEffect` calls `setValidCode(true)` which will cause a re-render, causing an infinite loop. Small bug unrelated to the ticket but I thought I'd fix it here
* Note the `TODO:`'s which should be resolved once the PUT endpoint is implemented

## Checklist

### DIFF
--- a/backend/typescript/rest/camperRoutes.ts
+++ b/backend/typescript/rest/camperRoutes.ts
@@ -101,6 +101,14 @@ camperRouter.get("/refund/:refundCode", async (req, res) => {
   }
 });
 
+camperRouter.put("/refund/:refundCode", async (req, res) => {
+  try {
+    res.status(200).json([]); // TODO: replace [] with proper response
+  } catch (error: unknown) {
+    res.status(500).json({ error: getErrorMessage(error) });
+  }
+});
+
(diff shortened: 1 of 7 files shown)


BraydenRoyston:
Looks good - won't be merging until we have the API work completed.

BraydenRoyston (approved):
Made some adjustments to integrate with our refund endpoint. Looking great overall.

